PR #8, opened 2025-04-07
Replace carbon foam support of VT planes with carbon fiber plates

fprino:
Hi Ruben,

I modified the VT geometry removing the carbon foam support for each plane and adding carbon fiber plates, with thickness of 400 um.
This way the geometry in the acceptance region should match the one of the fast simulation.
For the density of the carbon fiber, I used 1.91 g/cm3, which I took from the ITS geometry in O2.

I run the clang-formatting according to your instructions. I hope I did it right.

Thanks!

Cheers, Francesco

### DIFF
--- a/sim/src/NA6PVerTel.cxx
+++ b/sim/src/NA6PVerTel.cxx
@@ -97,19 +97,21 @@ void NA6PVerTel::createGeometry(TGeoVolume* world)
                                          param.shiftVerTel[2] + (param.posVerTelPlaneZ[0] + boxDZ / 2 - boxDZMargin),
                                          NA6PTGeoHelper::rotAroundVector(0.0, 0.0, 0.0, 0.0));
   world->AddNode(vtContainer, composeNonSensorVolID(0), vtTransform);
+
   // pixel station Frame with holes (box with subtracted holes)
-  auto* pixStFrameBox = new TGeoBBox("PixStFrameBox", frameDX / 2, frameDY / 2, frameDZ / 2);
-  auto* beamPipeHole = new TGeoTube("PixStFrameBoxBPHole", 0, frameHoleR, frameDZ);
-  auto* frameSubtraction = new TGeoSubtraction(pixStFrameBox, beamPipeHole);
-  auto* pixStFrameShape = new TGeoCompositeShape("PixStFrameBoxHole0", frameSubtraction);
-  for (size_t ii = 0; ii < chipHoleX.size(); ++ii) {
-    auto* pixChipHole = new TGeoBBox("PixChipHole", pixChipHoleDX / 2, pixChipHoleDY / 2, frameDZ);
-    auto* holeTransform = new TGeoTranslation(chipHoleX[ii], chipHoleY[ii], 0);
-    frameSubtraction = new TGeoSubtraction(pixStFrameShape, pixChipHole, nullptr, holeTransform);
-    pixStFrameShape = new TGeoCompositeShape(Form("PixStFrameBoxHole0%zu", ii), frameSubtraction);
-  }
-  TGeoVolume* pixStFrame = new TGeoVolume("PixStFrame", pixStFrameShape, NA6PTGeoHelper::instance().getMedium(addName("CarbonFoam")));
-  pixStFrame->SetLineColor(NA6PTGeoHelper::instance().getMediumColor(addName("CarbonFoam")));
+  // COMMENTED OUT: to be replaced with the new Al frame
+  // auto* pixStFrameBox = new TGeoBBox("PixStFrameBox", frameDX / 2, frameDY / 2, frameDZ / 2);
+  // auto* beamPipeHole = new TGeoTube("PixStFrameBoxBPHole", 0, frameHoleR, frameDZ);
+  // auto* frameSubtraction = new TGeoSubtraction(pixStFrameBox, beamPipeHole);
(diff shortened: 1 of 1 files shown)


sim/src/NA6PVerTel.cxx
```diff
@@ -107,23 +119,31 @@ void NA6PVerTel::createGeometry(TGeoVolume* world)
...
 
   // place sensors to station
   std::vector<float> alpdx{pixChipDX / 2 + pixChipOffsX, -pixChipDX / 2 + pixChipOffsY, -pixChipDX / 2 - pixChipOffsX, pixChipDX / 2 - pixChipOffsY};
   std::vector<float> alpdy{pixChipDY / 2 - pixChipOffsY, pixChipDY / 2 + pixChipOffsX, -pixChipDY / 2 + pixChipOffsY, -pixChipDY / 2 - pixChipOffsX};
   for (size_t ii = 0; ii < alpdx.size(); ++ii) {
     auto* sensorTransform = new TGeoTranslation(alpdx[ii], alpdy[ii], 0);
     pixelStationVol->AddNode(pixelSensor, composeSensorVolID(ii), sensorTransform);
+    int nns = pixelStationVol->GetNodes()->GetEntries() - 1;
+    auto* cbTransform = new TGeoTranslation(alpdx[ii], alpdy[ii], pixChipDz / 2 + carbonPlateDz / 2);
+    pixelStationVol->AddNode(cbPlate, composeNonSensorVolID(ii + 40), cbTransform);
```
shahor02: Don't we have a single carbon plate for the whole station? Here you are adding 1 plate per chip, i.e. 4 separate plates of the same size as the chip is.

sim/src/NA6PVerTel.cxx
```diff
@@ -107,23 +119,31 @@ void NA6PVerTel::createGeometry(TGeoVolume* world)
...
+    int nnc = pixelStationVol->GetNodes()->GetEntries() - 1;
+    LOGP(info, "Adding nodes {} and {} to {}, n. nodes = {}", pixelStationVol->GetNode(nns)->GetName(), pixelStationVol->GetNode(nnc)->GetName(), pixelStationVol->GetName(), pixelStationVol->GetNodes()->GetEntries());
```
shahor02: Could we hide this logging under 
```
if (mVerbosity > 0) {
  LOGP(info, ...
}
```
?

sim/src/NA6PVerTel.cxx
```diff
@@ -107,23 +119,31 @@ void NA6PVerTel::createGeometry(TGeoVolume* world)
...
   for (int ll = 0; ll < param.nVerTelPlanes; ++ll) {
     auto* stationTransform = new TGeoCombiTrans(param.posVerTelPlaneX[ll], param.posVerTelPlaneY[ll], param.posVerTelPlaneZ[ll] - zoffs,
                                                 NA6PTGeoHelper::rotAroundVector(0.0, 0.0, 0.0, 0.0));
     vtContainer->AddNode(pixelStationVol, composeNonSensorVolID(ll), stationTransform);
-    auto* frameTransform = new TGeoCombiTrans(param.posVerTelPlaneX[ll], param.posVerTelPlaneY[ll], param.posVerTelPlaneZ[ll] + 0.5 * (frameDZ + pixChipDz) - zoffs,
-                                              NA6PTGeoHelper::rotAroundVector(0, 0.0, 0.0, 0.0));
-    vtContainer->AddNode(pixStFrame, composeNonSensorVolID(ll + 20), frameTransform);
+    //    auto* frameTransform = new TGeoCombiTrans(param.posVerTelPlaneX[ll], param.posVerTelPlaneY[ll], param.posVerTelPlaneZ[ll] + 0.5 * (frameDZ + pixChipDz) - zoffs,
+    //                                              NA6PTGeoHelper::rotAroundVector(0, 0.0, 0.0, 0.0));
+    //    vtContainer->AddNode(pixStFrame, composeNonSensorVolID(ll + 20), frameTransform);
```
shahor02, line 148: The old pixStFrame volume is created above but it is never added. If it is not needed (see the comment above about the plate) then it should be removed from the code.

sim/src/NA6PVerTel.cxx
fprino: I chose to implement it like this (4 separate plates) for the time being in order to have in a simple way the internal hole matching the one of the chips. I can change to a single plate for the whole station if you think it is better.

sim/src/NA6PVerTel.cxx
fprino: Yes, sure. This was a remnant printout from the first tests, I can modify as you suggest in the next commit.

fprino:
This commit should address your suggestions.
Thanks for checking!
Cheers, Francesco

shahor02:
Thanks @fprino !

The whole frame https://github.com/shahor02/NA6PRoot/pull/8/files#diff-ce3bc9c2b3f374b0d4df51f7f4bef78cf13e8b7d8100b3930e2b6f898d0a34a1R101-R112 which before was made from C-foam but should be made from Al is now created but never added to the geometry. 
I would propose to comment it out at the moment. Eventually we should implement is as a proper Al frame holding 2 chips (once we have drawings).

fprino:
Thanks for the feedback.
I commented out the lines defining the old carbon-foam support.
I am planning to implement also the Al frame, but I postponed it since I don't have their specs.

shahor02:
Thanks!

PR #9, opened 2025-07-15
Fix overlap for carbon plate + fix typo in chip shifts

fprino:
Hi Ruben,
with this commit I modified the implementation of the carbon plate to remove an overlap with the  volume PixelStationVol.
While doing these checks, the overlap checker also spotted overlaps between pixel sensors:
PixelStationVol/PixelSensor_3100 overlapping PixelStationVol/PixelSensor_3101 ovlp=0.00246988
Checking in the code, it seems to me that this was due to a typo in the alpdx and alpdy arrays to define the offset.
With the modified code the overlap is no longer there.
Can you please check that what I did is correct?
Thank you!
Cheers, Francesco

### DIFF
--- a/sim/src/NA6PVerTel.cxx
+++ b/sim/src/NA6PVerTel.cxx
@@ -71,7 +71,7 @@ void NA6PVerTel::createGeometry(TGeoVolume* world)
 
   float pixChipContainerDX = frameDX;
   float pixChipContainerDY = frameDY;
-  float pixChipContainerDz = 0.05f;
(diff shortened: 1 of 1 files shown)


shahor02 (approved):
Hi @fprino ,
Thanks and sorry for the stupid error! Looks correct. 
@galocco, in case your ACTS geometry (showing smaller than fastsim effect of the C-plate),  was generated from the NA6Proot setup with this bug, this potentially can explain the reason of the discrepancy. Could you check?

PR #11, opened 2025-08-27
Add modular MS

galocco:
This PR add a modular muon spectrometer. The main motivation is that in ACTS, we need a central hole in all planes to split the setup into two volumes: one for the Vertex Spectrometer (VS) and one for the Muon Spectrometer (MS) (see [discussion](https://mattermost.web.cern.ch/acts/pl/wmpzhkdkm786zk3831trzbkm3y)
).
Additionally, the MS planes are designed to have different strip densities in different regions to cope with varying rates:
modularMS

### DIFF
--- a/base/include/NA6PLayoutParam.h
+++ b/base/include/NA6PLayoutParam.h
@@ -54,12 +54,14 @@ struct NA6PLayoutParam : public na6p::conf::ConfigurableParamHelper<NA6PLayoutPa
   float shiftMS[3] = {0.f, 0.f, 0.f}; // MS global shift, added to posMSPlaneX,Y,Z
   float posMSPlaneX[MaxMSPlanes] = {};
   float posMSPlaneY[MaxMSPlanes] = {};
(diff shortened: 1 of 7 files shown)


shahor02:
@galocco thanks, I cannot access this mattermost, but I believe that the change is needed :). Looks fine for me, but should not we introduce some gaps between the modules? Also, perhaps the modules size (chipDX, chipDY) and number of columns, rows should be added as parameters in the NA6PLayoutParam ?

galocco:
I added the sensor sizes as parameters in NA6PLayoutParam.
The number of columns and rows is now computed from the total plane size.

We should not add gaps, since the current plan is to stack the sensors to avoid dead zones.

The design shown in the figure is slightly outdated, and as far as I know, there is no final design yet. The latest plan is something like this:
station
However, this design cannot be directly converted into an ACTS geometry where the VS and MS are separate volumes. Splitting the detector into two volumes makes it simpler to reconstruct VS-only and MS-only tracks.

shahor02 (approved):
Thanks!

PR #24, opened 2025-11-17
Add option in setOrigin to properly treat different cases

fprino:


shahor02:
@fprino I've noticed that `mOriginSet` of generators was not reset after setting it in the 1st event. 
Is your PR meant to address this issue? If so, it was fixed 1h ago in my previous https://github.com/shahor02/NA6PRoot/pull/23: at the start of the new event the mOriginSet of all generators is reset.

fprino:
Yes, it was to fix this behaviour when SetOrigin is not imposed from the outside. Instead, when it is imposed from the outside it sets the same value for all events. If this is no longer needed, we can drop my commit

shahor02:
The `origin` was never supposed to be set permanently from outside, I had it to be able to set the primary vertex from the generator, and propagate it to the stack and other generators (e.g. in case of cocktail). And by mistake, it was never reset. 
Now it is reset, and also there is a possibility to provide a user macro with an arbitrary vertex generation algorithm (Enrico wanted a way to simulate events in the plug).  Could you check if you still see problems w/o your PR?

fprino:
Yes, I'll check. I had actually a use case in which I was calling setOrigin in my generator macro in order to set the vertex from the outside: this was done to force the collision to occur in 0,0,0 to have the same origin as in ACTS.

shahor02:
I see, I forgot that the mOrigineSet was added by you. As I wrote, with my last pr it will be anyway reset for every event.
Since now there is a possibility to fully customize the vertex generation using the macro, shall we simply remove it?

PR #30, opened 2025-12-03
First implementation of a NA6PRecoParam class

fprino:


base/include/NA6PRecoParam.h
```diff
@@ -0,0 +1,30 @@
+// NA6PCCopyright
```
shahor02: Should not this go to the rec directory? 
Also, since it is specifically for the CA, it would be better to reflect this in the name, in case later we have other reconstructors.

base/include/NA6PRecoParam.h
```diff
@@ -0,0 +1,30 @@
...
+#include "ConfigurableParamHelper.h"
+
+struct NA6PRecoParam : public na6p::conf::ConfigurableParamHelper<NA6PRecoParam> {
+
+  static constexpr int MaxIterationsTrackerCA = 10;
+
+  // VerTel reconstruction parameters
+  int    nVerTelLayers = 5;
+  int    nIterationsTrackerCA = 2;
+  double maxDeltaThetaTrackletsCA[MaxIterationsTrackerCA] = {0.04, 0.1, 0.15, 0.3, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0};
```
shahor02: Is there a reason to use double instead of float? This may cost extra conversions later. I would use doubles only where the precision is important, e.g. matrix manipulations.

base/include/NA6PRecoParam.h
fprino: I have no strong opinion. I put in base because I was copying the structure of the file na6PLayout.ini and I saw that the implementation of that class was in base. If you think it is better to move it into rec, this is perfectly fine with me

fprino: About the CA: actually I thought that this class will contain also the parameters of the clusterizer, the vertexer, of other trackers etc. This was just the first core and for the time being it contains only the parameters of the tracker CA, as a starting point. If you think it is better to split the files by "step" of the reconstruction I'll rename it.

base/include/NA6PRecoParam.h
fprino: Good point. I'll change to float

base/include/NA6PRecoParam.h
shahor02: Concerning the recoparam: I just wanted to avoid it becoming too bulky if we start adding different unrelated options in a single class. But ok, we can split it later if this become a problem.

For the directories: to avoid circular references, the base contains the stuff which both  sim and rec may depend on, the rec can depend on sim

shahor02 (approved):

PR #36, opened 2025-12-12
Add muon spectrometer reconstruction

galocco:
This PR introduces the NA6PMuonSpecReconstruction class to steer the muon spectrometer reconstruction.
MS and VT clusters are now represented by two dedicated classes, NA6PMuonSpecCluster and NA6PVerTelCluster, both inheriting from NA6PBaseCluster.
The cellular automaton and track fitter methods that previously relied on NA6PBaseCluster now use templates.
NA6PRecoParam has not yet been split into two; the current values correspond to the VT.

This PR also adds example macros to test the MS reconstruction (runMSTrackFinderCA.C, runMSTracking.C) and generalizes runFastFitOnHits.C.
Attached are some very preliminary efficiencies and resolutions for muons from J/ψ at √s = 8.8 GeV.

muon_resolutions
muon_purity
muon_imppar
muon_eff

### DIFF
--- a/rec/include/NA6PBaseCluster.h
+++ b/rec/include/NA6PBaseCluster.h
@@ -73,9 +73,9 @@ class NA6PBaseCluster
   float mSigYZ = 0.f;          // covariance matrix elements
   float mSigZZ = 0.f;          // covariance matrix elements
   int   mCluSiz = 0;           // cluster size
-  short mDetectorID = 0;       // the detector/sensor id
   int   mParticleID = 0;       // particle ID in Kine tree (MC truth)
   int   mHitID = -1;           // hit ID (for test of hitsToRecPoints)
+  short mDetectorID = 0;       // the detector/sensor id
   int8_t mLayer = -1;
 
   ClassDefNV(NA6PBaseCluster, 1);
--- a/rec/include/NA6PBaseCluster.h
+++ b/rec/include/NA6PBaseCluster.h
@@ -30,7 +30,8 @@ class NA6PBaseCluster
   NA6PBaseCluster& operator=(const NA6PBaseCluster&) = default;
   virtual ~NA6PBaseCluster() {}
 
-  virtual int getLayer() const { return -1; } // to be overridden in derived classes
+  int getLayer() const { return mLayer; }
+  int   getClusterSize() const { return mCluSiz; }
   // Lab frame coordinates
   auto getXLab()       const {return mPos[0];}
   auto getYLab()       const {return mPos[1];}
@@ -75,6 +76,7 @@ class NA6PBaseCluster
   short mDetectorID = 0;       // the detector/sensor id
   int   mParticleID = 0;       // particle ID in Kine tree (MC truth)
   int   mHitID = -1;           // hit ID (for test of hitsToRecPoints)
+  int8_t mLayer = -1;
 
   ClassDefNV(NA6PBaseCluster, 1);
 };
--- a/rec/include/NA6PBaseCluster.h
+++ b/rec/include/NA6PBaseCluster.h
@@ -30,6 +30,7 @@ class NA6PBaseCluster
   NA6PBaseCluster& operator=(const NA6PBaseCluster&) = default;
   virtual ~NA6PBaseCluster() {}
 
+  virtual int getLayer() const { return -1; } // to be overridden in derived classes
   // Lab frame coordinates
   auto getXLab()       const {return mPos[0];}
   auto getYLab()       const {return mPos[1];}
@@ -64,7 +65,6 @@ class NA6PBaseCluster
 
   virtual void print() const;
   std::string asString() const;
-
   
  protected:
   float mPos[3] = {};          // cartesian position of cluster in lab frame
--- a/rec/include/NA6PBaseCluster.h
+++ b/rec/include/NA6PBaseCluster.h
@@ -25,7 +25,7 @@ class NA6PBaseCluster
  public:
 
   NA6PBaseCluster() = default;
-  NA6PBaseCluster(float x, float y, float z, int clusiz);
+  NA6PBaseCluster(float x, float y, float z, int clusiz, int layer);
   NA6PBaseCluster(const NA6PBaseCluster&) = default;
   NA6PBaseCluster& operator=(const NA6PBaseCluster&) = default;
   virtual ~NA6PBaseCluster() {}
--- a/rec/include/NA6PFastTrackFitter.h
+++ b/rec/include/NA6PFastTrackFitter.h
@@ -17,12 +17,12 @@
 
 #include <string>
 #include <Rtypes.h>
+#include "NA6PBaseCluster.h"
 
 // Fast track fit based on Kalman filter
 
 class TGeoManager;
 class NA6PTrack;
-class NA6PBaseCluster;
 
 class NA6PFastTrackFitter
 {
@@ -35,7 +35,7 @@ class NA6PFastTrackFitter
 
   // setters for configurable parameters
   void   setMaxChi2Cl(double v=10)  {mMaxChi2Cl = v;}
-  void   setNLayersVT(int n) {
+  void   setNLayers(int n) {
     mNLayersVT = n; mClusters.clear(); mClusters.resize(n);
   }
   void   setParticleHypothesis(int pdg);
@@ -53,7 +53,8 @@ class NA6PFastTrackFitter
   void   computeSeed();
   void   printSeed() const;
   
-  void   addClusterVT(int jLay, NA6PBaseCluster* cl);
+  template <typename ClusterType>
+  void   addCluster(int jLay, ClusterType* cl);
   void   resetClusters() {
     // deletes the owned cluster and sets pointer to nullptr
     for (auto& clPtr : mClusters) clPtr.reset();
@@ -65,8 +66,9 @@ class NA6PFastTrackFitter
 
   bool   loadGeometry(const char* filename = "geometry.root", const char* geoname = "NA6P");
   
-  NA6PTrack*  fitTrackPointsVT();
-  bool   updateTrack(NA6PTrack* trc, NA6PBaseCluster* cl) const;
+  NA6PTrack*  fitTrackPoints();
+  template <typename ClusterType>
+  bool   updateTrack(NA6PTrack* trc, ClusterType* cl) const;
   int    propagateToZ(NA6PTrack* trc, double zFrom, double zTo, int dir) const;
   int    propagateToZ(NA6PTrack* trc, double zTo) const;
   void   getMeanMaterialBudgetFromGeom(double* start, double* end, double *mparam) const;
--- a/rec/include/NA6PFastTrackFitter.h
+++ b/rec/include/NA6PFastTrackFitter.h
@@ -36,7 +36,9 @@ class NA6PFastTrackFitter
   // setters for configurable parameters
   void   setMaxChi2Cl(double v=10)  {mMaxChi2Cl = v;}
   void   setNLayers(int n) {
-    mNLayers = n; mClusters.clear(); mClusters.resize(n);
+    mNLayers = n;
+    mClusters.clear();
+    mClusters.resize(n);
   }
   void   setParticleHypothesis(int pdg);
   void   enableMaterialCorrections() {mCorrectForMaterial = true;}
@@ -53,8 +55,7 @@ class NA6PFastTrackFitter
   void   computeSeed();
   void   printSeed() const;
   
-  template <typename ClusterType>
-  void   addCluster(int jLay, ClusterType* cl);
+  void   addCluster(int jLay, NA6PBaseCluster* cl);
   void   resetClusters() {
     // deletes the owned cluster and sets pointer to nullptr
     for (auto& clPtr : mClusters) clPtr.reset();
@@ -67,8 +68,7 @@ class NA6PFastTrackFitter
   bool   loadGeometry(const char* filename = "geometry.root", const char* geoname = "NA6P");
   
   NA6PTrack*  fitTrackPoints();
-  template <typename ClusterType>
-  bool   updateTrack(NA6PTrack* trc, ClusterType* cl) const;
+  bool   updateTrack(NA6PTrack* trc, NA6PBaseCluster* cl) const;
   int    propagateToZ(NA6PTrack* trc, double zFrom, double zTo, int dir) const;
   int    propagateToZ(NA6PTrack* trc, double zTo) const;
   void   getMeanMaterialBudgetFromGeom(double* start, double* end, double *mparam) const;
--- a/rec/include/NA6PFastTrackFitter.h
+++ b/rec/include/NA6PFastTrackFitter.h
@@ -36,7 +36,7 @@ class NA6PFastTrackFitter
   // setters for configurable parameters
   void   setMaxChi2Cl(double v=10)  {mMaxChi2Cl = v;}
   void   setNLayers(int n) {
-    mNLayersVT = n; mClusters.clear(); mClusters.resize(n);
+    mNLayers = n; mClusters.clear(); mClusters.resize(n);
   }
   void   setParticleHypothesis(int pdg);
   void   enableMaterialCorrections() {mCorrectForMaterial = true;}
@@ -81,9 +81,7 @@ class NA6PFastTrackFitter
   static const Double_t kAlmostZero;
 
  protected:
-  int  mNLayersVT = 5;                   // number of active VT layers in the model
-  int  mNLayersMS = 4;                   // number of active MS layers in the model
-  int  mNLayersTR = 2;                   // number of active Trigger layers in the model
+  int  mNLayers = 5;                   // number of active 
   double mMaxChi2Cl = 10.;               // max cluster-track chi2
   bool   mIsSeedSet = false;             // flag for set seed
   int    mSeedOption = kThreePointSeed;  // seed option (see enum)
@@ -102,7 +100,7 @@ class NA6PFastTrackFitter
   
   int getNumberOfClusters() const{
     int nClus = 0;
-    for (int jLay = 0; jLay < mNLayersVT; ++jLay) if(mClusters[jLay]) nClus++;
+    for (int jLay = 0; jLay < mNLayers; ++jLay) if(mClusters[jLay]) nClus++;
     return nClus;
   }
 
--- a/rec/include/NA6PMuonSpecCluster.h
+++ b/rec/include/NA6PMuonSpecCluster.h
@@ -24,13 +24,11 @@ class NA6PMuonSpecCluster : public NA6PBaseCluster
  public:
 
   NA6PMuonSpecCluster() = default;
-  NA6PMuonSpecCluster(float x, float y, float z, int clusiz) : NA6PBaseCluster(x, y, z, clusiz) {}
+  NA6PMuonSpecCluster(float x, float y, float z, int clusiz);
   NA6PMuonSpecCluster(const NA6PMuonSpecCluster&) = default;
   NA6PMuonSpecCluster& operator=(const NA6PMuonSpecCluster&) = default;
   virtual ~NA6PMuonSpecCluster() {}
-
-  int getLayer() const override;
-
+  
   ClassDefNV(NA6PMuonSpecCluster, 1);
 };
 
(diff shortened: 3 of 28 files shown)


rec/include/NA6PTrack.h
```diff
@@ -111,8 +108,7 @@ class NA6PTrack
  protected:
   double   mMass = 0.140;                        // particle mass
   double   mChi2 = 0.f;                          // total chi2
-  double   mChi2VT = 0.f;                        // total chi2 VT
-  int      mNVTLayers = 5;                       // number of vT laters
+  int      mNLayers = 6;                       // number of layers
```
fprino: Shouldn't the default here be 11 (5 layers of VT + 6 of MS?

rec/include/NA6PTrack.h
```diff
@@ -111,8 +108,7 @@ class NA6PTrack
  protected:
   double   mMass = 0.140;                        // particle mass
   double   mChi2 = 0.f;                          // total chi2
-  double   mChi2VT = 0.f;                        // total chi2 VT
```
fprino, line 114: I could be worth to keep separate the chi2 in the VT an the chi2 in the MS (in case we will use the same class also for the matched tracks)?

rec/src/NA6PTrack.cxx
```diff
@@ -229,7 +226,7 @@ std::string NA6PTrack::asString() const
   double pxyz[3];
   getPXYZ(pxyz);
   return fmt::format("Track: Nclusters:{} NVTclusters:{}  chi2:{} pos:{:.4f},{:.4f},{:.4f} mom:{:.3f},{:.3f},{:.3f}",
-                     mNClusters,mNClustersVT,mChi2,getXLab(),getYLab(),getZLab(),pxyz[0],pxyz[1],pxyz[2]);
+                     mNClusters,mNClusters,mChi2,getXLab(),getYLab(),getZLab(),pxyz[0],pxyz[1],pxyz[2]);
```
fprino: Here we should keep mNClustersVT, and maybe add mNClustersMS, which will be sued for the matched tracks

rec/src/NA6PTrack.cxx
```diff
@@ -243,18 +240,18 @@ void NA6PTrack::addCluster(const NA6PBaseCluster* clu, int cluIndex, double chi2
...
-  else if (nDet >= 24) mNClustersTR++;
+  
+  int nLay = clu->getLayer();
+  mClusterPartID[nLay] = trackID;
+  mClusterIndices[nLay] = cluIndex;
+  mClusterMap |= (1<<nLay);
+
+  // if (nDet < mNLayers * 4) {
+  // }
+  // else if (nDet >= 20 && nDet < 24) mNClustersMS++;
```
fprino: I would keep the logic to count separately clusters in VT, MC and possibly TR.

rec/src/NA6PVerTelCluster.cxx
```diff
@@ -0,0 +1,20 @@
+// NA6PCCopyright
+
+#include "NA6PVerTelCluster.h"
+#include "NA6PLayoutParam.h"
+
+int NA6PVerTelCluster::getLayer() const {
```
fprino: This implementation of getLayer is much slower than getSensorID()/4. Since we use it several times in the tracker, I would keep the fast and simple. An alternative option could be to store the layer in the cluster itself and fill it at the moment of the creation of clusters. This will increase a bit the size of the NA6PBaseCluster object adding 1 short, but it is probably acceptable.

rec/src/NA6PVerTelCluster.cxx
```diff
@@ -0,0 +1,20 @@
...
+	const float dzWindow = 2.5f; // tighter window for VT planes
+	for (int i = 0; i < layout.nVerTelPlanes; ++i) {
+		float zPlane = layout.posVerTelPlaneZ[i];
+		if (z > (zPlane - dzWindow) && z < (zPlane + dzWindow)) {
+			return i;
+		}
+	}
+	return -1;
+}
+// NA6PCCopyright
```
fprino, line 5: I guess lines 18-20 should be removed

rec/include/NA6PBaseCluster.h
```diff
@@ -30,6 +30,7 @@ class NA6PBaseCluster
   NA6PBaseCluster& operator=(const NA6PBaseCluster&) = default;
   virtual ~NA6PBaseCluster() {}
 
+  virtual int getLayer() const { return -1; } // to be overridden in derived classes
```
shahor02: Do we really need virtual methods which come at price? In most of use-cases a simple inheritance + overriding is enough, this is what we do in O2. 
And if the call has a virtual method, then one must use `ClassDef` for the CLING instead of `ClassDefNV`.

rec/include/NA6PMuonSpecCluster.h
```diff
@@ -0,0 +1,37 @@
...
+
+class NA6PMuonSpecCluster : public NA6PBaseCluster
+{
+ public:
+
+  NA6PMuonSpecCluster() = default;
+  NA6PMuonSpecCluster(float x, float y, float z, int clusiz) : NA6PBaseCluster(x, y, z, clusiz) {}
+  NA6PMuonSpecCluster(const NA6PMuonSpecCluster&) = default;
+  NA6PMuonSpecCluster& operator=(const NA6PMuonSpecCluster&) = default;
+  virtual ~NA6PMuonSpecCluster() {}
```
shahor02: In case we go on with virtual methods, in the derived classes they should be declared as e.g. 
```
~NA6PMuonSpecCluster() {} override;
```

rec/include/NA6PBaseCluster.h
fprino: When checking the code it also came into my mind (see one of the comments above) that an alternative (and maybe more economic) solution to treat the layer number could be to add a short mLayer as a data member of NA6PBaseCluster. This would avoid  the need of simple inheritance+overriding and also the usage of calculations of the number of layer in the tracking phase.

shahor02: Actually. even `int8_t` would be enough for the layer ID. That said, the layer can be extracted from the detectorID, we can add later a helper for that.
Also, allocating int hitID and particleID for the real clusters might be too much, especially taking into account that the cluster may correspond to >1 particle or hit. Eventually, I would simply port to NA6Proot the Alice O2 MCCompLabel and container.

fprino: We can indeed extract the layer from the detectorID. For the VT this is a division by 4, and it is whta is currently done in the code. However, Giacomo told me that for the MS the correspondence detector->layer is different. We could  implement a logic that based on the detector ID checks if it is a VT or MS clusters and returns the layer number. This is even more economic than adding an int8_t in the clusters to store the layer.
For the MC labels, the current implementation is indeed minimal, mostly a placeholder and works in the current situation in which we do hit->recpoint via smearing, so that 1 cluster corresponds to 1 hit. Using the MCCompLabel from O2 is a good solution.

galocco: Following your suggestion, I added an int8_t member to BaseCluster to store the layer and removed the virtual methods.
This implies also storing the layer for VT clusters, even though it is not strictly necessary.

For the MS, obtaining the layerId is currently less efficient than for the VT, since the number of sensors per plane varies, and the MS detectorId is an ambiguous identifier.
This limitation may be resolved once the MS geometry is fully defined.

rec/include/NA6PTrack.h
galocco: mNLayers and mNVTLayers were not used in the code, so I removed both.

rec/include/NA6PTrack.h
galocco: you're right, I added again mChi2VT and added mChi2MS

rec/src/NA6PVerTelCluster.cxx
shahor02: In principle, it is worth keeping the clusters of different layers in separate containers, so the algorithm should anyway know the layer of the cluster from the container it accesses.

rec/include/NA6PBaseCluster.h
```diff
@@ -75,6 +76,7 @@ class NA6PBaseCluster
   short mDetectorID = 0;       // the detector/sensor id
   int   mParticleID = 0;       // particle ID in Kine tree (MC truth)
   int   mHitID = -1;           // hit ID (for test of hitsToRecPoints)
+  int8_t mLayer = -1;
```
shahor02, line 79: Could you move the `short mDetectorID = 0;` before the `int8_t mLayer = -1;` ? Since the largest datamember has sizeof=4, with the current layout the mDetectorID (2bytes) will be padded to 4 bytes, the same for mLayer (1byte), leading to the waste of 4 bytes. With `mDetectorID` before the `mLayer` they will be stored in the same 4-bytes word and only 1 byte will be added for padding.

rec/include/NA6PFastTrackFitter.h
```diff
@@ -35,8 +35,8 @@ class NA6PFastTrackFitter
 
   // setters for configurable parameters
   void   setMaxChi2Cl(double v=10)  {mMaxChi2Cl = v;}
-  void   setNLayersVT(int n) {
-    mNLayersVT = n; mClusters.clear(); mClusters.resize(n);
+  void   setNLayers(int n) {
+    mNLayers = n; mClusters.clear(); mClusters.resize(n);
```
shahor02: We don't have yet strict coding conventions, but I would stick more or less to those of alice O2, let's avoid multiple commands on the same line. This reduces the readability and complicates debugging.

rec/include/NA6PMuonSpecCluster.h
shahor02: Were not you eliminating the virtuality?

rec/include/NA6PTrack.h
```diff
@@ -27,7 +27,7 @@ class NA6PTrack
 {
  public:
   
-  enum {kNDOF=5, kMaxVTLr=16};
+  enum {kNDOF=5, kMaxLr=16};
```
shahor02, line 30: We should probably set the exact number of layers

shahor02: ... but since the track should be eventually derived from the base track provided parameterization only this can be done later.

rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -46,13 +45,13 @@ NA6PFastTrackFitter::NA6PFastTrackFitter() :
   mSeedMom[0] = mSeedMom[1] = mSeedMom[2] = 1.; // 1 GeV for default momentum seed
 }
 
-
-void NA6PFastTrackFitter::addClusterVT(int jLay, NA6PBaseCluster* cl) {
-  if (jLay < 0 || jLay >= mNLayersVT) {
+template <typename ClusterType>
+void NA6PFastTrackFitter::addCluster(int jLay, ClusterType* cl) {
```
shahor02: Definition of templated method in the cxx file creates problems for compilers, better have them in the header file.
But I don't understand why do we need to templates here at all: you are storing the pointer to the cluster and all types of clusters anyway derive from the base cluster. You can simply pass the pointer on the base cluster.

Then, what is the reason to use std::unique_ptr (question probably to @fprino )? This will invalidate the original raw pointer once the unique pointer is destroyed.

rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -247,6 +247,15 @@ bool NA6PFastTrackFitter::updateTrack(NA6PTrack* trc, NA6PBaseCluster* cl) const
   return true;
 }
 
+// Explicit instantiations to ensure symbols are emitted in the library
+template void NA6PFastTrackFitter::addCluster<NA6PMuonSpecCluster>(int, NA6PMuonSpecCluster*);
+template void NA6PFastTrackFitter::addCluster<NA6PVerTelCluster>(int, NA6PVerTelCluster*);
+template void NA6PFastTrackFitter::addCluster<NA6PBaseCluster>(int, NA6PBaseCluster*);
```
shahor02: OK, this frees from the necessity to have the templated method definition in the header file, still, not clear why we need 3 methods, if the underlying pointer is anyway  `NA6PBaseCluster*`?

rec/src/NA6PMuonSpecCluster.cxx
```diff
@@ -0,0 +1,21 @@
+// NA6PCCopyright
+
+#include "NA6PMuonSpecCluster.h"
+#include "NA6PLayoutParam.h"
+#include <cmath>
+
+NA6PMuonSpecCluster::NA6PMuonSpecCluster(float x, float y, float z, int clusiz)
```
shahor02: Why not passing the layer ID as an argument?

rec/src/NA6PVertexerTracklets.cxx
```diff
@@ -114,15 +114,14 @@ void NA6PVertexerTracklets::printConfiguration() const
 
 //______________________________________________________________________
 
-void NA6PVertexerTracklets::sortClustersByLayerAndEta(std::vector<NA6PBaseCluster>& cluArr,
+void NA6PVertexerTracklets::sortClustersByLayerAndEta(std::vector<NA6PVerTelCluster>& cluArr,
```
shahor02, line 117: May need a separate PR, but I think it would be easier to have a vector of clusters per layer.

rec/src/NA6PVertexerTracklets.cxx
```diff
@@ -149,15 +148,15 @@ void NA6PVertexerTracklets::sortClustersByLayerAndEta(std::vector<NA6PBaseCluste
...
-      float r2b = xb * xb + yb * yb;
+    std::sort(first, last, [](const NA6PVerTelCluster& a, const NA6PVerTelCluster& b) {
+      double xa = a.getX();
+      double ya = a.getY();
+      double za = a.getZ();
+      double xb = b.getX();
+      double yb = b.getY();
+      double zb = b.getZ();
+      double r2a = xa * xa + ya * ya;
+      double r2b = xb * xb + yb * yb;
```
shahor02: why double?

fprino: This Is probably because Giacomo did not start from the latest versione of the vertexer class. Now, I have moved ti float everywhere

rec/src/NA6PVertexerTracklets.cxx
fprino: Yes! Indeed It would make sense to have the clusters already grouped per layer. For the first implementation I sticked to the strutture used for hits. Of we all agree I can take care of updating the tree of clusters in a next PR

rec/src/NA6PFastTrackFitter.cxx
fprino: Indeed my proposal of storing the layer number in the clusters was mainly meant to use the NA6PBaseCluster in the tracker and vertexer classes. For the std_unique_pointer, I'll check later

rec/include/NA6PTrack.h
fprino: I chose to set a max number of layer instrad of the actual one to allow for tests with varying number of layers, e.g. tests with additional TOF planes in the VT

rec/src/NA6PVertexerTracklets.cxx
shahor02: Ah, @galocco could you please rebase your PR to the freshly updated `main`?

rec/src/NA6PFastTrackFitter.cxx
fprino: I checked in the fitter the std::unique_ptr. The reason is that we have as data member a vector of pointers to the clusters to be fitted (mClusters). These clusters are passed to the fitter by the method addCluster. I wanted to avoid copies of the cluster objects. The smart pointer seemed to me a good solution to deal with the memory usage. An alternative approach could be to define the mCluster data member as a vector of pointers:
std::vector<const NA6PBaseCluster*> mClusters;
and change the add cluster method into:
void addClusterVT(int jLay, const NA6PBaseCluster& cl)
{
  if (jLay < 0 || jLay >= mNLayersVT) {
    LOGP(error,"Invalid layer index {}",jLay);
    return;
  }
  mClusters[jLay] = &cl;
}
What do you think?

rec/src/NA6PVertexerTracklets.cxx
galocco: My apologies. I did it, but I wasn’t careful while resolving the conflicts. I will fix it.

rec/src/NA6PFastTrackFitter.cxx
shahor02: I was wondering why the unique_ptr is better than the usual pointer? One could simply have 
```
std::vector<NA6PBaseCluster*> mClusters;
```
or better 
```
std::vector<const NA6PBaseCluster*> mClusters;
```
using  as a setter
```
void NA6PFastTrackFitter::addCluster(int jLay, ClusterType* cl) {  // or addCluster(int jLay, const ClusterType* cl) in the 2nt case
  if (jLay < 0 || jLay >= mNLayers) {
    LOGP(error,"Invalid layer index {}",jLay);
    return;
  }
  mClusters[jLay] = cl;
}
```

and then there is no risk of accidentally invalidating the original cluster pointer: having a unique_ptr and raw pointer  on the same object is very error-prone.


Rechecking the code I noticed other thing: in  https://github.com/shahor02/NA6PRoot/blob/3422059a0c9d4f3e61d78b2a2e077f7f5653cbdd/rec/src/NA6PTrackerCA.cxx#L473-L480 you loop over clusters propagating the track to given cluster and cutting on it track-cluster chi2. But you don't update the track when going from layer to layer, so the same initial state (at the 1st cluster is propagated). The selectivity of such a check will drop with the distance from the initial cluster since the cov. matrix always blows up.
Is this really what you want to do or the update was forgotten?

rec/src/NA6PMuonSpecCluster.cxx
galocco: If you mean the detectorID rather than the layer, then it is a good idea to pass it.
While the layer is not trivial to compute, so I preferred not to pass it explicitly.

rec/src/NA6PFastTrackFitter.cxx
galocco: You're right, the template can be removed. I'll update it.

rec/src/NA6PFastTrackFitter.cxx
fprino: Here is a possible modification to avoid the usage of unique_ptr: https://github.com/fprino/NA6PRoot/commit/e2b3c102d1180a13df0ae040d8eb823024e9ff13
Let me know if you think it is ok.
The diff file in the commit is large because I applied the clang format. The relevant differences are in the definition of mClusters, and in the methods resetClusters and addClustersVT of the fitter. I had also to adapt the trackerCA and the example macro accordingly.

fprino: I double checked the question of the track update in the clusters. Actually the loop at lines 473-480 of the trackerCA is a remnant of the first implementation that was using genfit in which I was propagating the track though the layers in order to compute the track-to-cluster chi2. With the fast fitter the cut on the chi2 is already done in NA6PFastTrackFitter::updateTrack and it is sufficient the line mTrackFitter->setMaxChi2Cl(maxChi2TrClu); to apply it. So, if I am not mistaken, I can simply remove the loop at lines 473-480. Ill run a test to make sure that this is the case. Thank you for spotting!

rec/src/NA6PMuonSpecCluster.cxx
shahor02: Actually, I meant the layer: the determination of the layer from Z depends on particular layout, while the layer of the cluster is known a priori.

galocco: ok, I will change the constructor to pass the layer explicitly.

The Z-based layer determination
  mDetectorID = nDet;
  // Set layer based on Z position and layout parameters
  float zCluster = getZLab();
  const auto& layout = NA6PLayoutParam::Instance();
  const float dzWindow = 20.f;
  for (int i = 0; i < layout.nMSPlanes; ++i) {
    float zPlane = layout.posMSPlaneZ[i];
    if (zCluster > (zPlane - dzWindow) && zCluster < (zPlane + dzWindow)) {
      mLayer = i + layout.nVerTelPlanes;
      return;
    }
  }

will be moved to convertHitsToRecPoints.C.
I will do the same for VT clusters.

shahor02 (commented):
Hi @galocco @fprino ,

Shall I merge it as is? I see there a still lot of doubles used when they are not really necessary, but this I can change later globally.

PR #38, opened 2025-12-15
Vertexer parameters in NA6PRecoParam + use xbeam and ybeam

fprino:
This adds the parameters of the vertexer to the NA6PRecoParam.
In addition there are a number of minor modifications and fixes to properly deal with the cases of beam shifted in x and y.
clang format applied to the modified files.

### DIFF
--- a/rec/include/NA6PTrackerCA.h
+++ b/rec/include/NA6PTrackerCA.h
@@ -39,7 +39,7 @@ struct CellCandidate {
   int startingLayer;
   int firstTrackletIndex;
   int secondTrackletIndex;
-  std::array<int,3> cluIDs;
+  std::array<int, 3> cluIDs;
   NA6PTrack trackFitFast;
   //    genfit::Track trackFit;
 };
@@ -52,7 +52,7 @@ struct TrackCandidate {
   std::vector<int> cluIDs;
 };
 
-struct TrackFitted{
+struct TrackFitted {
   int innerLayer;
   int outerLayer;
   int nClus;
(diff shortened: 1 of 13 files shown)


rec/include/NA6PTrackerCA.h
```diff
@@ -62,138 +62,129 @@ struct TrackFitted{
...
+                          double maxDeltaPhiTracklets,
+                          double maxDeltaTanLCells,
+                          double maxDeltaPhiCells,
+                          double maxDeltaPxPzCells,
+                          double maxDeltaPyPzCells,
+                          double maxChi2TrClCells,
+                          double maxChi2ndfCells,
+                          double maxChi2ndfTracks,
+                          int minNClusTracks);
   void configureFromRecoParam(const std::string filename = "");
```
shahor02: in principle, all const objects should be passed by reference.

rec/include/NA6PVertexerTracklets.h
```diff
@@ -114,6 +114,8 @@ class NA6PVertexerTracklets
   void setUseStandardKDE() { mKDEOption = kStandardKDE; }
   void setUseAdaptiveKDE() { mKDEOption = kAdaptiveKDE; }
   void setVerbosity(bool opt = true) { mVerbose = opt; }
+  void configureFromRecoParam(const std::string filename = "");
```
shahor02: reference?

rec/src/NA6PVertexerTracklets.cxx
```diff
@@ -13,6 +14,104 @@ ClassImp(NA6PVertexerTracklets)
   configurePeakFinding(mZMin, mZMax, mNBinsForPeakFind);
 }
 
+void NA6PVertexerTracklets::configureFromRecoParam(const std::string filename)
```
shahor02: reference?

rec/src/NA6PVertexerTracklets.cxx
```diff
@@ -266,9 +365,9 @@ void NA6PVertexerTracklets::computeIntersections(const std::vector<TrackletForVe
     switch (mRecoType) {
       case kYZ: {
         // work in the yz plane (non bending) to use straight line approximation of traks
```
shahor02, line 367: for the performance reason, the calculations done in a loop and which are not particularly precision-demaning, better to do in float.

fprino:
Thanks Ruben for the checks and the advices. I changed the arguments to use references in those methods.
Concerning the float vs double: I thought that deltay and deltax between clusters could be smallish and therefore the double precision was useful in this case. But maybe I am overdoing. I'll change to float and run a test

shahor02:
Hi Francesco, the smallness of the value per see does not necessarily require double prec., the float guarantees > 7 decimal digits, which are usually enough for what we are doing. Usually, the double are needed when round-off error is accumulated over many operations or when the operations over the numbers with a large difference in their orders of magnitude are involved. I don't think this is a case here.

fprino:
OK. I moved double->float everywhere in the vertexer. From a test there is actually no significant difference on the found vertex coordinates. I will then move from double to float in NA6PVertex

PR #39, opened 2025-12-18
Avoid usage of unique_ptr in the fitter

fprino:


fprino:
After some checks, I have removed the loop with the cut on the track-cluster chi2 in the trackerCA method that calls the fitter. This required to change the return value in NA6PFastTrackFitter::updateTrack in case the chi2 cut was not passed. The "return true" which was in the original code was taken from na60fastsim, but actually to me it seems correct that the fit is abandoned in this case rather than just skipping the track update with the cluster. Does this make sense? I can use a different logic otherwise.

rec/include/NA6PFastTrackFitter.h
```diff
@@ -28,49 +28,60 @@ class NA6PFastTrackFitter
 {
 
  public:
-  enum {kTwoPointSeed = 0,kThreePointSeed = 1};
+  enum { kTwoPointSeed = 0,
+         kThreePointSeed = 1 };
 
   NA6PFastTrackFitter();
-  virtual ~NA6PFastTrackFitter() {};
+  virtual ~NA6PFastTrackFitter(){};
```
shahor02: looks like correct changes by the clang-format were undone. More important: just noticed that the method is declared virtual (why, are we going to inherit from this class) but the class is using ClassDefNV.
Root reserves ClassDefNV only for non-virtual classes, otherwise ClassDef must be used.

rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -161,202 +180,216 @@ void NA6PFastTrackFitter::getMeanMaterialBudgetFromGeom(double* start, double* e
...
   double meas[2] = {cl->getYTF(), cl->getZTF()}; // ideal cluster coordinate, tracking (AliExtTrParam frame)
   double measErr2[3] = {cl->getSigYY(), cl->getSigYZ(), cl->getSigZZ()};
 
-  double chi2 = trc->getTrackExtParam().getPredictedChi2(meas,measErr2);
-  if (chi2>mMaxChi2Cl) return true; // chi2 is too large
-    
-  if (!trc->update(meas,measErr2)) {
+  double chi2 = trc->getTrackExtParam().getPredictedChi2(meas, measErr2);
+  if (chi2 > mMaxChi2Cl)
+    return false; // chi2 is too large
```
shahor02, line 264: Agree, I think this is a correct behaviour)

fprino:
This last commit should have resolved the conflicts with respect to PR #36. I applied the clang format via clang-format - before committing. I also removed the virtual from the destructor of the fast fitter class

fprino:
I am wondering if we need the ClusterType template in NA6PTrackerCA. I thought that we could simply use NA6PBaseCluster everywhere, as done in the fitter. @galocco , @shahor02 : what is your take on this?

shahor02:
Thanks @fprino ! For the template / NA6PBaseCluster: in principle, the reconstruction in the MS, once the strips are introduced, may require extra cluster methods absent in the pixel clusters (then we should see if the CA tracking is really good for MS).
But I would prefer to get rid of the current NA6PBaseCluster with the fast sim. trick legacy, which also propagated to the NA6P... tracking code, e.g. usage of getYTF etc. 
I would propose that you push/merge all your developments asap then freeze your activity for 1-2 days, so that I could rebase everything to native cluster/track classes. 
I will have time for that tomorrow and half of 24th, do you have something else to commit? @galocco ?

fprino:
Ok. Thanks! Your proposal of moving to your native params is perfecly fine with me. The only development I have open Is that in PR 40, which I will conclude this evening, so that tomorrow you will be free to do your modifs and rebase everything

PR #40, opened 2025-12-22
Add vertexer algo based on tracklet pairs

fprino:


fprino:
This commit adds an alternative vertex reconstruction algorithm which implements the tracklet-pair based vertexer from O2 ITS reconstruction. It differs from the existing methods because: 1) it doesn't include a peak finding approach, but it computes the intersections of all tracklet pairs and clusterizes them based on distance selections; 2) it provides 3D (x,y,z) determination of the vertex position.

rec/include/NA6PVertex.h
```diff
@@ -32,12 +32,14 @@ class NA6PVertex
...
 
-  enum vertTypes { kTrackletPrimaryVertex,
+  enum vertTypes { kTrackletPrimaryVertexZ,
+                   kTrackletPrimaryVertex3D,
                    kTrackPrimaryVertex,
                    kBaseVertex };
 
   NA6PVertex() = default;
   NA6PVertex(const ROOT::Math::XYZPointF& pos, const std::array<float, kNCov>& cov, int nCont, float chi2);
+  NA6PVertex(const std::array<float, 3> xyz, int nCont);
```
shahor02: should be passed by reference, `const std::array<float, 3>& xyz`

shahor02 (commented):
Hi @fprino , @galocco ,
Just 1 comment below, but also https://github.com/shahor02/NA6PRoot/pull/36 has lot of modifications to NA6PVerTelReconstruction and conflicts with this PR. Shall I 1st merge PR36 then we resolve the conflicts or vice versa?

fprino:
Hi Ruben! 
Yes, please, go ahead and merge PR 36. 
I will then take care of updating this PR resolving the conflicts and modifying also the pass by reference of xyz as you suggested.

shahor02:
OK, done

fprino:
But there are still other conflicts that break the compilation, due to the usage of VerTel vs Base clusters. I'll debug more later and commit the relevant changes

shahor02:
For me a simple change
```
diff --git a/rec/include/NA6PVertexerTracklets.h b/rec/include/NA6PVertexerTracklets.h
index f477c8a..b440657 100644
--- a/rec/include/NA6PVertexerTracklets.h
+++ b/rec/include/NA6PVertexerTracklets.h
@@ -197,7 +197,7 @@ class NA6PVertexerTracklets
   bool findVertexKDE(const std::vector<TracklIntersection>& zIntersec,
                      std::vector<NA6PVertex>& vertices);
   bool compute3DVertices(const std::vector<TrackletForVertex>& selTracklets,
-                         const std::vector<NA6PBaseCluster>& cluArr,
+                         const std::vector<NA6PVerTelCluster>& cluArr,
                          std::vector<NA6PVertex>& vertices);
   void printStats(const std::vector<TrackletForVertex>& candidates,
                   const std::vector<NA6PVerTelCluster>& cluArr,
diff --git a/rec/src/NA6PVertexerTracklets.cxx b/rec/src/NA6PVertexerTracklets.cxx
index 19fc057..ea4f1a1 100644
--- a/rec/src/NA6PVertexerTracklets.cxx
+++ b/rec/src/NA6PVertexerTracklets.cxx
@@ -734,7 +734,7 @@ bool NA6PVertexerTracklets::findVertexKDE(const std::vector<TracklIntersection>&
 //_______________________________________________________________________
 
 bool NA6PVertexerTracklets::compute3DVertices(const std::vector<TrackletForVertex>& selTracklets,
-                                              const std::vector<NA6PBaseCluster>& cluArr,
+                                              const std::vector<NA6PVerTelCluster>& cluArr,
                                               std::vector<NA6PVertex>& vertices)
 {
   const int nTracklets = selTracklets.size();
```
makes your PR to compile with the main

fprino:
Thanks. Yes, that should be enough. But I wanted to try also to check if we could stay with the base cluster in the vertexer class for simplicity + retest everything After update all the code. I'll do after dinner and update the PR

fprino:
Given the time constraints, I kept the VerTel cluster. i'll check again after you will have merged the native parameterization if we can use with the base class in the vertexer.
With the last commit, the code compiles, I also moved the constructor to use a reference, as you suggested.
I run some basic tests and everything behaves as expected.

shahor02:
Thank, for the validation, could you outline the set of commands to use to validate your tracking starting from the simulation?

fprino:
For quick validation of the tracker, I use the macro NA6PRoot/test/runVTTrackFinderCA.C, which produces some basic plots that I compare to a reference reconstruction.
This macro runs directly the tracker on the clusters without using the interface of NA6PVerTelReconstruction.
If you want to test also the VerTelReconstruction class, you can use runVTTracking.C which produces a root file of tracks.
For the vertexer, I use runVertexerTracklets.C.

shahor02:
Hi @fprino 
I am now testing https://github.com/shahor02/NA6PRoot/pull/43.
Running runVertexerTracklets.C (with main branch) over 30 pythia pp events I get 0 vertices. Also I see 
[FATAL] Failed to read INI config file test2.ini (test2.ini: cannot open file)
If this file is necessary, could you please commit it?

fprino:
Hi Ruben,
Actually the file Is not necessary. The line in the macro was intended to show how to pass an external configuration file. But It should have been commented out. My mistake in preparing the code for commit. Let me know if you want me to commit an updated version.

shahor02:
Thanks Francesco,
But why do I see nothing in Pythia pp @ 150 GeV? Is it supposed to work with pp?

fprino:
I never tried it on pp events. If you can share the root files of hits and mckine I can have a look.

shahor02:
Hi @fprino ,
I put the tarball with pythia events to https://cernbox.cern.ch/files/spaces/eos/user/s/shahoian/tmp/savPPP30.tar.gz

fprino:
Hi @shahor02,
I do not manage to download the tar file. When I try to open the link I am redirected automatically to my home directory on cernbox, while if I try with "save link as" it downloads a small file which cannot be unpacked with tar xvfz.

shahor02:
Hi, it was probably not in the public dir, I have reloaded it to the Google drive

https://drive.google.com/file/d/16pbPX_NX48wKd1EreCVunyoaoi9mGE4d/view?usp=drivesdk

fprino:
Hi Ruben,
thanks. I downloaded the file and I understood why you did not find vertices. There are essentially two main reasons:
1) running the vertexer through NA6PVertTelReconstruction is missing the possibility to set the X and Y coordinates of the beam. In case of displacement of the vertex position in the transverse plane this affects the vertex finding efficiency for the methods using only the Z vertex
2) there was a request in the default configuration that the peak should contain minimum 3 intersections among tracklets, and this was killing the efficiency for low multiplicity events

I could retrieve some vertices (for 12-14 out of 20 events) with these configurations:
1) use the macro inspectVertexerTracklets.C (which does not go though NA6PVertTelReconstruction and sets the X and Y position of the beam) together with the configuration file testpp-KDE.ini (which uses the peak-finding Z-only method but lowers to request of tracklets in the peak to 2)
2) use the macro runVertexerTracklets.C together with the configuration file testpp-pairalgo.ini, which uses the ALICE O2 vertexer that computes the position on the vertex in 3D and does not need xbeam ad ybeam

Here is a link to a tgz file with the modified version of inspectVertexerTracklets.C (that should work with your branch) and the two ini files:
https://cernbox.cern.ch/s/dBETx1YeJLoPizC
I do not commit them now to avoid cluttering the development branch. 
Once your changes will be merged, I will take care of updating all the macros and also implement the usage of X and Y position of the beam when running the vertexer through NA6PVerTelReconstruction.

shahor02:
Thanks @fprino !

Will check once I am back.

Cheers 
  Ruben

PR #54, opened 2026-01-30
Add VT MS matching

galocco:
This PR adds a class for the VT MS matching, and test/runMatching.C provides an example of its use.
The matching chi2 is computed using the mOuter parametrization of VT tracks. To propagate VT and MS tracks to a common z position, the propagation logic in both the track fitter and the track classes has been updated.
The generic chi2 variable in NA6PTrack has been removed, as it was ambiguous with respect to chi2VT and chi2MS, and replaced by mMatchChi2.

### DIFF
--- a/rec/include/NA6PMatching.h
+++ b/rec/include/NA6PMatching.h
@@ -52,25 +52,21 @@ class NA6PMatching : public NA6PReconstruction
   void setVerTelTracks(const std::vector<NA6PTrack>& tracks)
   {
     mVerTelTracks = tracks;
-    hVerTelTrackPtr = &mVerTelTracks;
   }
 
   void setMuonSpecTracks(const std::vector<NA6PTrack>& tracks)
   {
     mMuonSpecTracks = tracks;
-    hMuonSpecTrackPtr = &mMuonSpecTracks;
   }
 
   void setVerTelClusters(const std::vector<NA6PVerTelCluster>& clusters)
   {
     mVerTelClusters = clusters;
-    hVerTelClusPtr = &mVerTelClusters;
   }
 
   void setMuonSpecClusters(const std::vector<NA6PMuonSpecCluster>& clusters)
   {
     mMuonSpecClusters = clusters;
-    hMuonSpecClusPtr = &mMuonSpecClusters;
   }
 
   bool init(const char* filename, const char* geoname = "NA6P") override;
--- a/rec/include/NA6PMatching.h
+++ b/rec/include/NA6PMatching.h
@@ -1,0 +1,130 @@
+// NA6PCCopyright
+
+// Minimal matching reconstruction driver
+
+#ifndef NA6P_MATCHING_H
+#define NA6P_MATCHING_H
+
+#include <string>
+#include <Rtypes.h>
+#include <TVector3.h>
+
+#include "NA6PReconstruction.h"
+#include "NA6PVerTelCluster.h"
+#include "NA6PMuonSpecCluster.h"
+#include "NA6PFastTrackFitter.h"
+
+#include "NA6PTrack.h"
+#include "NA6PVertex.h"
+class TH1D;
+class TH2D;
+
+class TFile;
+class TTree;
+class NA6PFastTrackFitter;
+
+class NA6PMatching : public NA6PReconstruction
+{
+ public:
+  NA6PMatching();
+  ~NA6PMatching() override = default;
+
+  void useMCMatching() { mMCMatching = true; }
+  void useChi2Matching() { mMCMatching = false; }
+
+  void setMaxChi2Match(double v) { mMaxChi2Match = v; }
+  void setMaxChi2Refit(double v) { mMaxChi2Refit = v; }
+  void setMinTrackP(double p) { mMinTrackP = p; }
+  void setMinVTHits(int n) { mMinVTHits = n; }
+  void setMinMSHits(int n) { mMinMSHits = n; }
+  void setPMatchWindow(double w) { mPMatchWindow = w; }
+  void setZMatching(double z)
+  {
+    mZMatching = z;
+    mIsZMatchingSet = true;
+  }
+
+  void setPrimaryVertexPosition(double x, double y, double z)
+  {
+    mPrimaryVertex.SetXYZ(x, y, z);
+  }
+
+  void setVerTelTracks(const std::vector<NA6PTrack>& tracks)
+  {
+    mVerTelTracks = tracks;
+    hVerTelTrackPtr = &mVerTelTracks;
+  }
+
+  void setMuonSpecTracks(const std::vector<NA6PTrack>& tracks)
+  {
+    mMuonSpecTracks = tracks;
+    hMuonSpecTrackPtr = &mMuonSpecTracks;
+  }
+
+  void setVerTelClusters(const std::vector<NA6PVerTelCluster>& clusters)
+  {
+    mVerTelClusters = clusters;
+    hVerTelClusPtr = &mVerTelClusters;
+  }
+
+  void setMuonSpecClusters(const std::vector<NA6PMuonSpecCluster>& clusters)
+  {
+    mMuonSpecClusters = clusters;
+    hMuonSpecClusPtr = &mMuonSpecClusters;
+  }
+
+  bool init(const char* filename, const char* geoname = "NA6P") override;
+
+  void createTracksOutput() override;
+  void clearTracks() override
+  {
+    mMatchedTracks.clear();
+  }
+  void writeTracks() override;
+  void closeTracksOutput() override;
+  void sortVTTracksByP(std::vector<NA6PTrack>& tracks);
+  double computeChi2(const double* par1, const double* cov1,
+                     const double* par2, const double* cov2);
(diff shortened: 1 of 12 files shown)


rec/include/NA6PTrackerCA.h
```diff
@@ -193,7 +193,7 @@ class NA6PTrackerCA
   bool mVerbose = false;
   bool mPropagateTracksToPrimaryVertex = false;
   bool mDoOutwardPropagation = false;
-  float mZOutProp = 40.;
+  float mZOutProp = 38.1175;
```
fprino, line 196: Ultimately, I would not rely on this default value, but rather use setZForOutwardPropagation from NA6PMSReconstructor, possibly using a value taken from the geometry or from the reco param

rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -220,38 +220,76 @@ int NA6PFastTrackFitter::propagateToZ(NA6PTrack* trc, double zTo) const
...
     return -1;
-  if (std::abs(trc->getZLab() - zFrom) > 1.e-4) {
-    LOGP(fatal, "Track is not in the expected z position: {} vs {}", zFrom, trc->getZLab());
+  }
+  
+  // Select appropriate methods based on parametrization
+  auto getPosition = outer ? &NA6PTrack::getXYZOuter : &NA6PTrack::getXYZ;
+  auto getZ = outer ? &NA6PTrack::getZLabOuter : &NA6PTrack::getZLab;
+  
+  double currentZ = (trc->*getZ)();
```
fprino: Here I would suggest to avoid defining getPosition and getZ, but stick to something simpler like:
double currentZ = outer? trc->getZLabOuter() : trc->getZlab();
and then
double start[3];
if (outer) trc->getXYZOuter(start);
else trc->getXYZ(start);

rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -220,38 +220,76 @@ int NA6PFastTrackFitter::propagateToZ(NA6PTrack* trc, double zTo) const
...
+  (trc->*getPosition)(start);
+  
+  // Create temporary track and propagate without material to get endpoint
   NA6PTrack tmpTrc = *trc;
-  tmpTrc.propagateToZBxByBz(zTo);
-  tmpTrc.getXYZ(end);
-  double p1[3], p2[3];
-  trc->getXYZ(p1);
-  tmpTrc.getXYZ(p2);
+  if (!tmpTrc.propagateToZBxByBz(zTo, 1., 0., 0., outer)) {
```
fprino, line 258: Here to compute the end point we propagate the inward track parameterization also in the "outer" case. I guess it does not make a big difference for the estimation of the material budget.

rec/include/NA6PTrackerCA.h
galocco: I agree. Soon, I will open another PR to include the MS parameter for the CA in NA6PRecoParam. I can also add this parameter.

rec/src/NA6PFastTrackFitter.cxx
galocco: Following your suggestion, I simplified [NA6PTrack::propagateToZBxByBz](https://github.com/shahor02/NA6PRoot/pull/54/changes/d093edd05d0c8df7242891db7300163a50c0e322#diff-c5a98e482c1da5b0e29103a8b5627b90d57a66efa56669f0b343be837909386dR109) too

rec/src/NA6PFastTrackFitter.cxx
galocco: No, if outer is true, the endpoint is computed with mOuter

rec/include/NA6PMatching.h
```diff
@@ -0,0 +1,130 @@
...
+
+  void setPrimaryVertexPosition(double x, double y, double z)
+  {
+    mPrimaryVertex.SetXYZ(x, y, z);
+  }
+
+  void setVerTelTracks(const std::vector<NA6PTrack>& tracks)
+  {
+    mVerTelTracks = tracks;
+    hVerTelTrackPtr = &mVerTelTracks;
```
shahor02: since the hVerTelTrackPtr already points to mVerTelTracks address, there is no need to assign it again

rec/include/NA6PTrack.h
```diff
@@ -85,19 +88,22 @@ class NA6PTrack
...
+  double getPOuter() const { return mOuter.getP(); }
   double getSigmaX2() const { return mExtTrack.getSigmaY2(); }
   double getSigmaY2() const { return mExtTrack.getSigmaZ2(); }
   double getSigmaXY() const { return mExtTrack.getSigmaZY(); }
   double getSigmaP2() const;
   double getSigmaPX2() const;
   double getSigmaPY2() const;
   double getSigmaPZ2() const;
-  double getNormChi2() const { return mNClusters < 3 ? 0 : mChi2 / ((mNClusters << 1) - kNDOF); }
+  double getNormChi2() const { return mNClusters < 3 ? 0 : (mChi2VT+mChi2MS) / ((mNClusters << 1) - kNDOF); }
```
shahor02, line 102: This getNormChi2 can be misleading: the sum of individual fits in the ITS and MS does not amount for the chi2 of the global fit (which should indeed be distributed as chi2 for `mNClusters*2 - 5` DOFs).

rec/include/NA6PTrack.h
```diff
@@ -122,8 +128,9 @@ class NA6PTrack
   {
     return mExtTrack.correctForMeanMaterial(xOverX0, xTimesRho, mMass, anglecorr);
   }
-  bool propagateToZBxByBz(double z, double maxDZ = 1.0, double xOverX0 = 0., double xTimesRho = 0.);
+  bool propagateToZBxByBz(double z, double maxDZ = 1.0, double xOverX0 = 0., double xTimesRho = 0., bool outer = false);
   bool propagateToZBxByBz(double z, const double* bxyz) { return mExtTrack.propagateToBxByBz(z, bxyz); }
+  bool propagateToZBxByBzOuter(double z, const double* bxyz) { return mOuter.propagateToBxByBz(z, bxyz); }
   bool propagateToDCA(NA6PTrack* partner);
```
shahor02, line 134: All these propagate.. methods will become part of the standalone (static) propagator class in the PR I am still changing (like O2 Propagator). It will also provide the query for the material budget and mag.field.

rec/src/NA6PMatching.cxx
```diff
@@ -0,0 +1,279 @@
...
+{
+  std::sort(tracks.begin(), tracks.end(),
+            [](const NA6PTrack& a, const NA6PTrack& b) {
+              return a.getP() > b.getP();
+            });
+}
+
+double NA6PMatching::computeChi2(const double* par1, const double* cov1,
+                                 const double* par2, const double* cov2)
+{
```
shahor02, line 73: The calculation of chi2 between 2 tracks will be a method of the NA6PTrackParCov (equivalent of Alice ExternalTrackParam).

rec/include/NA6PMatching.h
galocco: I see. I just want to be sure, we don’t need it here ([link](https://github.com/galocco/NA6PRoot/blob/main/rec/include/NA6PVerTelReconstruction.h#L69)) or in the other places where this code was copied?

rec/include/NA6PTrack.h
galocco: mChi2VT and mChi2MS are not the chi2 of individual fits. They are the contributions to the chi2 from VT clusters and MS clusters, respectively, from the same fit.

rec/include/NA6PMatching.h
shahor02: For all cases: if you have `hXXXPtr = &mXXX;` in the class definition, you don't need to repeat this assignment in the setters for mXXX: the assignments will be done in the class constructor.

rec/include/NA6PTrack.h
shahor02: Sorry, missed that.

rec/include/NA6PTrack.h
galocco: For now, would you like me to keep them as they are? same for computeChi2

shahor02: Sure, I commented just to avoid you polishing the parts which will anyway change.

shahor02:
Is this ready to be merged?

galocco:
I believe it’s ready to be merged.

shahor02:
@galocco ok, but it has conflicts, probably after last commits of @fprino . Could you please rebaise to main and force-push?

rec/include/NA6PTrackerCA.h
fprino: After thinking better on this, I think that this Z value should go in the VT Reco Param, since it is applied in the VT tracking, instead of the MS reco param. I can add it in the VT reco param after this commit is merged

rec/src/NA6PFastTrackFitter.cxx
fprino: Right, I overlooked the outer in the arguments

galocco:
@shahor02, now it should be ready to be merged.

PR #57, opened 2026-02-04
Update RecoParam

galocco:
Add the MS Cellular Automaton and matching parameters to NA6PRecoParam.

### DIFF
--- a/rec/include/NA6PRecoParam.h
+++ b/rec/include/NA6PRecoParam.h
@@ -11,7 +11,7 @@ struct NA6PRecoParam : public na6p::conf::ConfigurableParamHelper<NA6PRecoParam>
   static constexpr int MaxIterationsTrackerCA = 10;
 
   // VerTel reconstruction parameters
-  int nLayers = 5;
+  int vtNLayers = 5;
   // tracklet vertexer
   int vertexerLayerToStart = 0;
   float vertexerMaxDeltaThetaTracklet = 0.6;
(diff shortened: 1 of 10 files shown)


shahor02 (approved):
Fine for me. Just as a comment:
1) in principle, one can use directly the recoparam settings in the `if...s` instead of transferring the recoparam values to the datamembers with a similar name.
2) Concerning the pMatchWindow parameter: since the momentum resolution is roughly proportional to the momentum, a simple bracket would be too lax or too strict. It would  safer to use a window in 1/p  (which is more or less normally distributed, while the error on p is very much skewed).

Will merge if @fprino has no other comments.

rec/include/NA6PRecoParam.h
```diff
@@ -39,19 +39,43 @@ struct NA6PRecoParam : public na6p::conf::ConfigurableParamHelper<NA6PRecoParam>
...
-  float maxDeltaPxPzCellsCA[MaxIterationsTrackerCA] = {0.02, 0.05, 0.1, 0.2, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0};
-  float maxDeltaPyPzCellsCA[MaxIterationsTrackerCA] = {2e-3, 7.5e-3, 1.5e-2, 3.e-2, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0};
-  float maxChi2TrClCellsCA[MaxIterationsTrackerCA] = {100., 500., 999.0, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0};
-  float maxChi2ndfCellsCA[MaxIterationsTrackerCA] = {100., 500., 999.0, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0};
-  float maxChi2ndfTracksCA[MaxIterationsTrackerCA] = {100., 500., 999.0, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0, 999.0};
-  int minNClusTracksCA[MaxIterationsTrackerCA] = {5, 3, 0, 0, 0, 0, 0, 0, 0, 0};
-
+  // VerTel CA tracker
+  bool doOutwardPropagation = false;
+  float zOutProp = 40.;
```
fprino: Maybe, to make the variable names more "user friendly", we could add vtTracker in front of doOutwardPropagation and zOutProp?

fprino:
I just posted a minor comment on variable names. Can also be considered for a next round of updates. I am fine to merge. Thanks!

rec/include/NA6PRecoParam.h
galocco: Good idea. I also added msNLayers and renamed the vt parameters.

PR #66, opened 2026-02-12
Options for usage of B field in seed calculation, needed for MS

fprino:


shahor02:
Hi @fprino 
Does current way of extracting the B-field for seeding pose a problem? I am asking, because eventually the field value is irrelevant: the Kalman operates with curvature which we extract from the seed directly, regardless the field value. The track parameter is q/p_perp, at the seeding point, so the seed should be assigned `curvature/(-0.299792458e-3*B)` with B at the 1st point.

fprino:
Hi Ruben,
for the muon spectrometer, in which the B field has strong variations along the seed length I see a substantial effect.
These are the momentum residuals for the seed calculated from the stations 1, 2, and 3 when I use the B at the middle point:
SeedResiduals321_Bmid
While these are the residuals when I use the integral:
SeedResiduals321_Bint


The Kalman fit in the 4 stations of the MS does not correct this wrong seed (I also tried enlarging the covariance matrix element for 1/pt as you suggested in the meeting):
MomResidKalman4Lay-Bmid

The determination of pz improved from this situation with the B at mid point:
momRes4CluSeedBmid
To this one with the integral of B:
momRes4CluSeedBint

shahor02:
Hi Francesco,

Thanks for the details. I think this means that we have to change the way the seeding is done for the MS, i.e. to extract the q/pPerp from the global via 3 points fit accounting for the real field profile in Z.
I suspect that if you plot the same differences multiplied by track q (or do them separately for mu+ and mu-) you will still see a bias. 
I'll do the math later, at the merging as is, since the plots for Bmean are obviously better than for Bmiddle.

PR #69, opened 2026-02-17
Adding scripts to run in parallel PYTHIA and na6psim with HEPMC

lucamicheletti93:
- Adding scripts to run in parallel PYTHIA and na6psim with HEPMC
- Update of runVertexTracklets.C and fix of GenPYTHIA.cxx


test/runSimHepMC.sh
```diff
@@ -0,0 +1,18 @@
+NJOBS=20
+NCORES=20
+NEVENTSPERJOB=100
+OUTDIR=real_config/PYTHIA_PbPb_MB
+LAYOUT=na6pLayout_real.ini
+GENERATOR=genHEPMCWithSpectTracking.C+
+SEEDSTART=0
+SEEDEND=$(($SEEDSTART + $NJOBS))
+EVENTSTOSKIP=$(($SEEDSTART))
+
```
shahor02, line 15: Could you add a check if the parralel is installed, e.g. 
```
if [[ -z $(which parallel) ]] ; then 
  echo "parallel is not available, install it first"
  exit 1
fi
```

lucamicheletti93: @shahor02  thanks for the suggestion! I have added the lines you suggested.

shahor02 (approved):

PR #71, opened 2026-02-18
Fix GDML and force top volume voxelization

galocco:
The voxelization of the GDML geometry was likely failing, which caused navigation and track propagation issues. As a result, particles were observed to remain in the world volume or become stuck inside the DipoleVT yokes.

Additionally, there was a geometry inconsistency: the top volume defined in BothMagnets.gdml was smaller than the corresponding NA60+/DiCE world volume, while the DipoleMS volume was oversized.


shahor02 (approved):
Thanks!

PR #73, opened 2026-02-19
Optimize tracklet vertexing

galocco:
selectTracklets had some unused variables and redundant checks, which have been cleaned up. The search range for upper_bound/partition_point has been narrowed to start from lower instead of the beginning of the collection.


galocco:
@fprino can you confirm that I'm not removing stuff that was meant to be used in the future?

fprino:
Hi Giacomo. Very nice. Indeed the cluArr was a "remnants" of the zero-th implementation of the tracklet selection method and I forgot to clean it up. Also the optimizations of the range and of the break in the loop look very good.

PR #74, opened 2026-02-20
Adding script and macros for embedding signal in background events

scompar:
This commit consists of 3 files. The script embed.sh for each of the NEVENTS_BCK background events (for example an Angantyr Pb-Pb events) runs na6psim NEVENTS_SIG times, with a signal generator to be supplied and with the embed.C macro ensuring that the signal events are generated at the same position of the primary vertex. On completion, MergeHits_embed.C is invoked. It merges the kinematics file and the hits file taking into account the needed offsets for this operation. After all the  NEVENTS_BCK events are used, the output trees for kinematics and hits are merged via hadd. The output mixed trees have been thoroughly tested and look correct.

### DIFF
--- a/test/embedding/MergeHits_embed.C
+++ b/test/embedding/MergeHits_embed.C
@@ -1,0 +1,282 @@
+#include <TSystem.h>
+#include <TFile.h>
+#include <TTree.h>
+#include <TParticle.h>
+#include "NA6PMCEventHeader.h"
+#include "NA6PMCGenHeader.h"
+#include "NA6PMuonSpecModularHit.h"
+#include "NA6PVerTelHit.h"
+
+
+void MergeHits_embed(int bckEvent,
+                     const char *fKS = "MCKine.root",
+                     const char *fKB = "MCKine_bck.root",
+                     const char *fHSM = "HitsMuonSpecModular.root",
+                     const char *fHBM = "HitsMuonSpecModular_bck.root",
+                     const char *fHSV = "HitsVerTel.root",
+                     const char *fHBV = "HitsVerTel_bck.root"
+		     )
+{
+  // Background event under study
+  const char* evStr = gSystem->Getenv("BKG_EVENT");
+  bckEvent = evStr ? atoi(evStr) : 0;
+
+  // ===============================
+  // Files
+  // ===============================
+  TFile* fKSig = TFile::Open(fKS);
+  TFile* fKBkg = TFile::Open(fKB);
+  TFile* fKOut = TFile::Open(Form("MCKine_mix_%d.root", bckEvent),"RECREATE");
+  
+  TFile* fHSigM = TFile::Open(fHSM);
+  TFile* fHBkgM = TFile::Open(fHBM);
+  TFile* fHOutM = TFile::Open(Form("HitsMuonSpecModular_mix_%d.root", bckEvent), "RECREATE");
+
+  TFile* fHSigV = TFile::Open(fHSV);
+  TFile* fHBkgV = TFile::Open(fHBV);
+  TFile* fHOutV = TFile::Open(Form("HitsVerTel_mix_%d.root", bckEvent), "RECREATE");
+
+  // ===============================
+  // Trees
+  // ===============================
+  TTree* tSigKin  = (TTree*)fKSig->Get("mckine");
+  TTree* tSigHitM = (TTree*)fHSigM->Get("hitsMuonSpecModular");
+  TTree* tSigHitV = (TTree*)fHSigV->Get("hitsVerTel");
+
+  TTree* tBkgKin  = (TTree*)fKBkg->Get("mckine");
+  TTree* tBkgHitM = (TTree*)fHBkgM->Get("hitsMuonSpecModular");
+  TTree* tBkgHitV = (TTree*)fHBkgV->Get("hitsVerTel");
+
+  // Clone structure ONLY
+  fKOut->cd();
+  TTree* tOutKin = new TTree("mckine", "mixed kinematics");
+  std::vector<TParticle>* mixmcArr = new std::vector<TParticle>();
+  tOutKin->Branch("tracks", &mixmcArr);
+  NA6PMCEventHeader* mixmcHead = nullptr;
+  tOutKin->Branch("header", &mixmcHead);
+
+  fHOutM->cd();
+  TTree* tOutHitM = tSigHitM->CloneTree(0); 
+  std::vector<NA6PMuonSpecModularHit>* mixHitsM = new std::vector<NA6PMuonSpecModularHit>();
+  tOutHitM->SetBranchAddress("MuonSpecModular", &mixHitsM);
+
+  fHOutV->cd();
+  TTree* tOutHitV = tSigHitV->CloneTree(0);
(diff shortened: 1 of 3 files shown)


test/embedding/embed.C
```diff
@@ -0,0 +1,28 @@
...
+{
+  const char* evStr = gSystem->Getenv("BKG_EVENT");
+  int iEvent = evStr ? atoi(evStr) : 0;
+
+  TFile* fKBkg = TFile::Open("MCKine_bck.root");
+  TTree* tBkgKin = (TTree*)fKBkg->Get("mckine");
+
+  NA6PMCEventHeader* bkgmcHead = nullptr;
+  tBkgKin->SetBranchAddress("header", &bkgmcHead);
+  tBkgKin->GetEntry(iEvent);
```
shahor02: You could avoid opening/closing the file for every event by using 
```
static TFile* fKBkg = TFile::Open("MCKine_bck.root");
static TTree* tBkgKin = (TTree*)fKBkg->Get("mckine");
static NA6PMCEventHeader* bkgmcHead = nullptr;

if (!bkgmcHead) {
  tBkgKin->SetBranchAddress("header", &bkgmcHead);
}
tBkgKin->GetEntry(iEvent);

```

test/embedding/embed.C
```diff
@@ -0,0 +1,28 @@
...
+
+  x = bkgmcHead->getVX();
+  y = bkgmcHead->getVY();
+  z = bkgmcHead->getVZ();
+
+  std::cout << "Using background event " << iEvent << std::endl;
+  std::cout << "with vertex at x " << x << " y " << y << " z " << z << endl;
+
+  fKBkg->Close();
```
shahor02: Then, this should be removed.

test/embedding/MergeHits_embed.C
```diff
@@ -0,0 +1,282 @@
...
+    mixmcHead->setVY(VYs);
+    mixmcHead->setVZ(VZs);
+    mixmcHead->setNTracks(NTrackss+NTracksb); // total number of tracks for mixed event
+    mixmcHead->setNPrimaries(NPrimariess+NPrimariesb); // total number of primaries for mixed event
+
+// add generator headers for signal and background
+    for(int ighs = 0; ighs < (int)sigmcHead->getNGenHeaders(); ighs++){
+      const NA6PMCGenHeader& genHeadsig = sigmcHead->getGenHeader(ighs); // 
+      cout << " Sig gen header " << ighs << " primary offset signal " <<  genHeadsig.getPrimariesOffset() << " secondary offset signal " << genHeadsig.getSecondariesOffset()  << " info " << genHeadsig.getInfo() << endl;
+      if (genHeadsig.getNPrimaries() !=0 && genHeadsig.getNSecondaries() !=0) mixmcHead->addGenHeader(genHeadsig);
```
shahor02, line 145: Why do you need to veto on Nprimaries and Nsecondaries? I guess, if the header was added to the original file, it should also go to the mixed one.

test/embedding/MergeHits_embed.C
```diff
@@ -0,0 +1,282 @@
...
+   }
+    for(int ighb = 0; ighb < (int)bkgmcHead->getNGenHeaders(); ighb++){
+      const NA6PMCGenHeader& genHeadbck = bkgmcHead->getGenHeader(ighb); // 
+      cout << " Bck gen header " << ighb << " primary offset signal " <<  genHeadbck.getPrimariesOffset() << " secondary offset signal " << genHeadbck.getSecondariesOffset()  << " info " << genHeadbck.getInfo() << endl;
+      if (genHeadbck.getNPrimaries() !=0 && genHeadbck.getNSecondaries() !=0) mixmcHead->addGenHeader(genHeadbck);
```
shahor02, line 150: idem.

test/embedding/MergeHits_embed.C
```diff
@@ -0,0 +1,282 @@
...
+    int nbckhitM=0;
+    int nbckhitV=0;
+    
+    for (const auto& h : *bkgHitsM) {
+      NA6PMuonSpecModularHit hit = h;
+      hit.setTrackID(hit.getTrackID() + offset); // Adjust reference to track number
+      mixHitsM->push_back(hit);
+//      cout << " Background event " << bckEvent << " Bck Hit Muon " << nbckhitM << "  x = " << hit.getX() << "  y = " << h.getY() <<"  z = " << h.getZ() << " ref. mixed track n. " << hit.getTrackID() << endl;
+      nbckhitM++;
+    }
```
shahor02: There is no need for intermediate hit, e.g.
```
for (const auto& hit : *bkgHitsM) {
   mixHitsM->emplace_back(hit)->setTrackID(hit.getTrackID() + offset);
}
```
same for the ITS

test/embedding/embed.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
```
shahor02, line 2: It is worh adding here a comment telling that the background simulation is expected to have been done in advance, with its output renamed to XXX_bck.root

shahor02 (commented):
Hi @scompar 
thanks, looks good. Please see some minor comments below.

test/embedding/embed.C
scompar: OK, implemented

test/embedding/embed.C
scompar: Yes, done

test/embedding/embed.sh
scompar: OK, added

shahor02: Thanks @scompar , but did you forget to push the updates?

test/embedding/MergeHits_embed.C
scompar: When reading back background and signal generator headers, in case either background or signal is generated via NA6PGenCocktail an extra header appears with 0 primaries and 0 secondaries. In the case below the signal is generated as a "cocktail" of J/psi only and one gets the output below

Processing MergeHits_embed_v2.C(0)...
 Bck Event 0 header 0 primary offset background 0 secondary offset background 164 info genHEPMC_x164
 Sig Event 0 header 0 primary offset signal 0 secondary offset signal 4294967295 info cocktail
 Sig Event 0 header 1 primary offset signal 0 secondary offset signal 1 info Jpsi_pdg443x1

It seems to me this specific header does not carry useful information, so I decided to remove it from the merged event.

test/embedding/MergeHits_embed.C
scompar: Same as comment above on line R145

test/embedding/MergeHits_embed.C
scompar: OK, done for both

scompar: P.S.: modified to mixHitsM->emplace_back(hit).setTrackID(hit.getTrackID() + offset); to avoid compilation error

test/embedding/embed.sh
scompar: Was just checking the output, now done!

shahor02 (approved):

PR #75, opened 2026-02-20
MS tracks: fit outward + match to MID + refit inward

fprino:
With this commit, the macro runMSTrackMIDTrackletMatching.C produces an output root file with a TTree of tracks that are built starting from 4-cluster tracks in the MS + matching to segments in the MID + outward propagation of the tracks to the MID (attaching the clusters in the MID) + inward refit of a 6-cluster track.
This should allow us to compare the performance (efficiency, fakes, resolutions ...) of 4-cluster (MS only) tracks, 6-clusters tracks (MS+MID reconstructed together), and 4+2 tracks (with this matching approach).
Note that the cuts used in the track-tracklet matching are preliminary and hard coded in the macro. They are just meant for initial debug and studies. If this approach will turn out to be the best one, I will do a better implementation in NA6PMuonSpecReconstruction.

### DIFF
--- a/rec/src/NA6PFastTrackFitter.cxx
+++ b/rec/src/NA6PFastTrackFitter.cxx
@@ -708,7 +708,9 @@ NA6PTrack* NA6PFastTrackFitter::fitTrackPoints(int dir, NA6PTrack* seed)
     zCurr = zNext;
   }
 
-  if (!isGoodFit)
+  if (!isGoodFit) {
+    delete currTr;
     return nullptr;
+  }
   return currTr;
 }
--- a/rec/src/NA6PFastTrackFitter.cxx
+++ b/rec/src/NA6PFastTrackFitter.cxx
@@ -56,6 +56,17 @@ void NA6PFastTrackFitter::addCluster(int jLay, const NA6PBaseCluster& cl)
   mClusters[jLay] = &cl;
 }
 
+void NA6PFastTrackFitter::printClusters() const
+{
+  LOGP(info, "N layers = {}", mNLayers);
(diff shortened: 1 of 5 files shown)


rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -633,11 +645,10 @@ NA6PTrack* NA6PFastTrackFitter::fitTrackPoints(int dir, NA6PTrack* seed)
       else
         computeSeedOuter();
     }
-    if (!mIsSeedSet)
-      LOGP(warn, "Track seed not computed properly, will run the fit w/o seed");
-    currTr = new NA6PTrack();
     if (mIsSeedSet)
       currTr->init(mSeedPos, mSeedMom, mCharge);
+    else
+      LOGP(warn, "Track seed not computed properly, will run the fit w/o seed");
```
shahor02, line 651: I've noticed that in the end of this method we return nullptr in case of the faile fit, w/o destroying the created track. This creates a memory leak.  Could you add in the end
```
if (!isGoodFit) {
   delete currTr;
   return nullptr;
}
```

fprino:
Hi Ruben,
Thanks for spotting the leak! I have committed the fix you suggested.
I would then refactor this method to avoid allocating memory on the heap, e.g. by changing the signature to 
bool NA6PFastTrackFitter::fitTrackPoints(NA6PTrack& trackToFit, int dir, NA6PTrack* seed)
Since this will impact on several files that call this method, I will implement this change in another PR.

shahor02:
Thanks, this indeed would be better.

PR #76, opened 2026-02-21
Avoid memory allocation on the heap in fitTrackPoints

fprino:


fprino:
These changes have the additional benefit of speeding up significantly the tracking.
With these modifications, track finding + fitting in the VT for 1000 Angantyr MB Pb-Pb collisions went down from 195 seconds to 110 seconds.

PR #84, opened 2026-02-27
Avoid cluster/track copies and add cluster index assignment in matching

galocco:
Implement the cluster indices introduced in https://github.com/shahor02/NA6PRoot/commit/f254b0967509cf54d8e225bd2533e40dbab4c96d .

### DIFF
--- a/rec/include/NA6PMatching.h
+++ b/rec/include/NA6PMatching.h
@@ -59,15 +59,8 @@ class NA6PMatching : public NA6PReconstruction
     mMuonSpecTracks = tracks;
   }
 
-  void setVerTelClusters(const std::vector<NA6PVerTelCluster>& clusters)
-  {
-    mVerTelClusters = clusters;
-  }
-
-  void setMuonSpecClusters(const std::vector<NA6PMuonSpecCluster>& clusters)
-  {
-    mMuonSpecClusters = clusters;
-  }
+  void setVerTelClusters(const std::vector<NA6PVerTelCluster>& clusters);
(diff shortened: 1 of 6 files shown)


shahor02:
I've just realized that the NA6PMatching does a full copy for setClusters. Do we need this? Why not simply having a pointer to which external clusters vector pointer  is assigned with setClusters (or read in in case of input from the tree)?
I you want a vector like access w/o dereferencing (i.e. (*ptr)[i]) for every access, we can have an extra span data member, i.e. 
```
std::span<NA6PMuonSpecCluster> mMSClusSpan{};
NA6PMuonSpecCluster* mMSClusPtr = nullptr;
...
void setMuonSpecClusters(std::vector<NA6PMuonSpecCluster>& clusters)
{
  mMSClusPtr = &clusters;
  mMSClusSpan = std::span<NA6PMuonSpecCluster>(clusters);
  ...
}
```

galocco:
Thank you for pointing this out. Indeed, copying the clusters (and tracks) is unnecessary. The same pattern is used in the VerTel and MuonSpec reconstruction. In those cases, one might worry that the sorting in the CA tracker breaks the correspondence between cluster indices and their positions in the std::vector. However, even there it is not an issue, because in NA6PRec we first loop over events for VT and MS reconstruction, and then perform a second loop for matching, reading the clusters from the TTree at each event.

I implemented your suggestion. Regarding std::span: I would not add it just to avoid dereferencing, so I didn't include it, but I can do it if you prefer.

shahor02 (approved):

PR #87, opened 2026-03-04
Use material-dependent Bethe-Bloch; add configurable geometry interrogation step

galocco:
The reconstructed J/ψ mass showed a 40 MeV shift when using the MS. This bias was caused by two factors: first, the Bethe-Bloch formula used for energy-loss correction employed silicon-specific parameters; second, the material recording step when propagating the track from zFrom to zTo was not splitted in steps, no matter the distance.
This PR splits the material interrogation into finer steps, controlled by the new configurable maxStepForMaterialRecording, and uses the mean Z, Z/A, and density to assign the correct Bethe-Bloch parameters.
With a finer step and proper Bethe-Bloch parameters, the J/ψ mass shift is corrected.

### DIFF
--- a/rec/src/NA6PFastTrackFitter.cxx
+++ b/rec/src/NA6PFastTrackFitter.cxx
@@ -242,6 +242,7 @@ int NA6PFastTrackFitter::propagateToZOuter(NA6PTrack* trc, double zTo) const
 
 int NA6PFastTrackFitter::propagateToZImpl(NA6PTrack* trc, double zFrom, double zTo, int dir, bool outer) const
 {
+  double maxStepForProp = 1;
   // Validate track state
   if (trc->getTrackExtParam().getAlpha() < 0) {
     return -1;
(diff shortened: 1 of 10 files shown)


rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -258,40 +259,53 @@ int NA6PFastTrackFitter::propagateToZImpl(NA6PTrack* trc, double zFrom, double z
...
 
-  // Calculate material budget along path
-  double mparam[7] = {0.};
-  if (mCorrectForMaterial) {
-    getMeanMaterialBudgetFromGeom(start, end, mparam);
-  }
+    // Create temporary track and propagate without material to get endpoint
+    NA6PTrack tmpTrc = *trc;
+    if (!tmpTrc.propagateToZBxByBz(zNext, 1., 0., 0., outer)) {
+      return 0;
```
shahor02, line 281: It is very expensive to create and propagated the full track with cov.matrix just to know its end point. But I am not sure it is even needed. Instead of getting straight-line material estimate for the large step (so it is certainly biased) then splitting it to small steps in `propagateToZBxByBz(zNext, 1., x2x0, -dir * xrho * corrELoss, outer, density, atomicZ, zOverA)`, it is better to do the propagation itself in a small steps with material query at every step. This is automatically done in the propagator for the track native implementation I am preparing (hope to finish this weekend): https://github.com/shahor02/NA6PRoot/blob/20d3da3589510940a6c7da2f97dee3430445d25b/rec/src/Propagator.cxx#L134

rec/src/NA6PFastTrackFitter.cxx
```diff
@@ -258,40 +259,53 @@ int NA6PFastTrackFitter::propagateToZImpl(NA6PTrack* trc, double zFrom, double z
...
+
+    double density = mparam[0];
+    double xrho = density * mparam[4]; // density * length
+    double x2x0 = mparam[1];             // X/X0
+    double atomicZ = mparam[3];
+    double zOverA = mparam[5];
+
+    double corrELoss = 1.0;    
+    // Propagate with material corrections
+    if (!trc->propagateToZBxByBz(zNext, 1., x2x0, -dir * xrho * corrELoss, outer, density, atomicZ, zOverA)) {
```
shahor02, line 305: In fact, this correction is still too crude as it does not account for the dedx change with the corrected momentum. The "native" track has more elaborate multi-step correction, https://github.com/shahor02/NA6PRoot/blob/20d3da3589510940a6c7da2f97dee3430445d25b/rec/src/NA6PTrackParCov.cxx#L624
though it still uses fixed BB parameters. 
What is the numeric difference between the dedx estimated accounting for Z, A and the current version?

galocco: ratioBB
These are the values of the dedx for most of the material that we can cross. On the bottom you can find the ration wrt the Si using the current bb function.

fprino:
Hi Giacomo, 
did you estimate separately the effects of the correction in the Bethe-Bloch parameters and the one of using a finer steps? 
I was doing tests by using a small step in NA6PFastTrackFitter::propagateToZ and the effect I saw on single muon momentum bias (and resolution) was smallish.
Cheers, Francesco

rec/src/NA6PFastTrackFitter.cxx
fprino: I had tried the propagation in small steps with a simple wrapper method:

```
int NA6PFastTrackFitter::propagateToZSteps(NA6PTrack* trc, double zTo, double step) const
{
  double zCurr = trc->getZLab();
  double length = std::abs(zTo - zCurr);
  int dir = (zTo - zCurr) > 0 ? 1 : -1;
  int nSteps = 1;
  if (length > step)
    nSteps = std::ceil(length/step);
  double zNext = zCurr;
  bool done = false;
  for (int iStep = 0 ; iStep < nSteps; ++iStep) {
    zNext += step * dir;
    if((dir == -1 && zNext < zTo) ||
       (dir == 1 && zNext > zTo)) {
      zNext = zTo;
      done = true;
    }
    if (!propagateToZImpl(trc, zCurr, zNext, dir, false))
      return 0;
    zCurr = zNext;
    if (done) break;
  }
  return 1;
}

```
and the effect on the single muon momentum bias was smallish. I thought it is because the tracks are close to straight lines in the absorber.

shahor02: In this case the point of having small steps would be accounting the material "gradually", since eloss is not linear with the material traversed.

rec/src/NA6PFastTrackFitter.cxx
shahor02: OK, thanks, for the materials in the absorber acceptance, it is a few % effect. I'll need to add this feature to the track param.

galocco:
Ciao Francesco, yes. I'm preparing a presentation that I can link here this afternoon. What I found is that both the step for the material interrogation and the Bethe-Bloch parameters contribute to momentum bias.
I also tried to reduce the step for trc->propagateToZBxByBz, but without improvement.

shahor02 (approved):
@galocco did not manage to work on na6p... this WE, so merging as is, will transfer relevant part to the propagator later

PR #91, opened 2026-03-10
First implementation of a track vertexer

fprino:
This PR ports the PVertexer class used to compute the primary vertex in O2. 
Following some indications from Ruben, I removed the time "dimension" from the vertex search.
I did not yet implement the step of debris reduction.
I did some basic checks on few Pb-Pb events, mainly to check that the code works and finds reasonable positions.

### DIFF
--- a/rec/include/NA6PVertex.h
+++ b/rec/include/NA6PVertex.h
@@ -43,7 +43,7 @@ class NA6PVertex
   NA6PVertex(const float* xyz, int nCont);
   NA6PVertex(const NA6PVertex&) = default;
   NA6PVertex& operator=(const NA6PVertex&) = default;
-  virtual ~NA6PVertex() = default;
+  ~NA6PVertex() = default;
 
   void init(const float* xyz, const float* cov, int nCont, float chi2);
   void setX(float x) { mPos.SetX(x); }
@@ -103,7 +103,7 @@ class NA6PVertex
   void addTrackID(int id) { mTrackIDs.push_back(id); }
   const std::vector<int>& getTrackIDs() const { return mTrackIDs; }
   
-  virtual void print() const;
+  void print() const;
   std::string asString() const;
 
  protected:
--- a/rec/include/NA6PVertex.h
+++ b/rec/include/NA6PVertex.h
@@ -100,15 +100,19 @@ class NA6PVertex
   float getChi2() const { return mChi2; }
   short getVertexType() const { return mVertexType; }
 
+  void addTrackID(int id) { mTrackIDs.push_back(id); }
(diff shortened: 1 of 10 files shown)


fprino:
This is the outcome from a first basic test on 100 Pb-Pb events (NA49 primaries) at 40 GeV:
PrimVertContrib

PrimVertexResiduals

rec/include/NA6PVertex.h
```diff
@@ -100,15 +100,19 @@ class NA6PVertex
   float getChi2() const { return mChi2; }
   short getVertexType() const { return mVertexType; }
 
+  void addTrackID(int id) { mTrackIDs.push_back(id); }
+  const std::vector<int>& getTrackIDs() const { return mTrackIDs; }
+  
   virtual void print() const;
```
shahor02: did not notice before: the class with ClassDefNV declaration should not have virtual methods. If such methods are needed, then one should use ClassDef(. But in this case there is no need for virtuality.

rec/include/NA6PVertexerTracks.h
```diff
@@ -0,0 +1,235 @@
...
+    mSigXYI = -sxy * detI;
+    // z weight from error propagation of transverse cov along track direction
+    // xz and yz correlations neglected (mSigXZI = mSigYZI = 0)
+    mSig2ZI = (szz > 0.f) ? 1.f / szz : 0.f;
+  }
+  bool isValid() const { return mSig2ZI > 0.f; }
+  bool canUse() const { return vtxID == kNoVtx; }
+  bool canAssign() const { return wgh > 0. && vtxID == kNoVtx; }
+
+  std::array<float, 3> getResiduals(const float vtxPos[3]) const
```
shahor02: `const float& vtxPos` ?

rec/include/NA6PVertexerTracks.h
```diff
@@ -0,0 +1,235 @@
...
+  {
+    // vector from closest point on line to vtxPos
+    auto comps = mLine.getDCAComponents(vtxPos);
+    return {comps[0], comps[3], comps[5]}; // dx, dy, dz components
+  }
+
+  float evalChi2ToVertex(const float vtxPos[3]) const
```
shahor02: same

shahor02 (commented):
Hi @fprino ,

Looks good! Just a few insignificant comments below.

fprino:
Hi Ruben,
thank you for reviewing the code. I have removed the virtual methods from the vertex, and changed to a const reference to a srd::array as the argument of getResiduals and evalChi2.

shahor02 (approved):
Thanks!

PR #99, opened 2026-03-22
Lot of changes to rebase to native NA6PTrackPar / NA6PTrackParCov

shahor02:
Not tested yet, questionable parts are flagged by RSTODO

@fprino @galocco I am almost done with the changes, though I have not yet tested the reconstruction (I expect it to fail, don't try yet).  I've moved as much functionality as possible to the track class itself and dedicate Propagator class
(in the end, the Alice like dipole-oriented parametrization with q/pxz appears to be more precise than the generic with 1/p, although, if we decide to have generic field Kalman filter, I can add converter for the alternative generic track model https://github.com/shahor02/NA6PRoot/blob/pr_nativetrF/rec/include/NA6PTrackPar.h)

Before testing, I still want to modify further the NA6PFastTrackFitter, if this is ok for you:
you set to it the full set of clusters to use, and failure to update with one of them amounts to a failed fit. But it operates on the full NA6PTrack (with all its alternative parametrizations, flags, cluster indices etc.), though the only thing it does apart from fitting is registering the clusters to the track. 
For this reason, doing forward/backward fits requires cloning a lot of unnecessary information instead of using the base track parametrizations as outwRefit and inwRefit objects e.g. in https://github.com/shahor02/NA6PRoot/blob/b5f5bfe3bcccf0af077df6e59c1d0cad7ec90644/rec/src/NA6PTrackerCA.cxx#L799-L802
Registering the clusters in the final full track can be delegated to a later stage in a dedicated method.
Also, there is some inconsisteny between the NA6PFastTrackFitter doing the propagation to the primary vertex and doing it optionally again e.g. in the https://github.com/shahor02/NA6PRoot/blob/b5f5bfe3bcccf0af077df6e59c1d0cad7ec90644/rec/src/NA6PTrackerCA.cxx#L821 or in the matching. 
Let me know if you prefer to discuss this on zoom, or I proceed on my own.


rec/src/Propagator.cxx
```diff
@@ -0,0 +1,265 @@
...
+          res = false;
+        }
+      }
+      return res;
+    };
+    if (opt.linRef) {
+      if (!(opt.byOnly ? track.propagateToZ(z, getBy(xyz0), linRef) : track.propagateToZ(z, getFieldXYZ(xyz0), linRef))) {
+        return false;
+      }
+    }
```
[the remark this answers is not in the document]
shahor02: fixed

rec/src/NA6PMatching.cxx
```diff
@@ -263,21 +259,12 @@ bool NA6PMatching::fitAndStoreMatchedTrack(const NA6PTrack& vtTrk,
...
     NA6PTrack vtSeed = vtTrk;
-
-  const float zFirstVTCluster =
-      (*hVerTelClusPtr)[vtTrk.getClusterIndex(0)].getZLab();
-
-    if (vtSeed.getZLab() != zFirstVTCluster) {
-      mTrackFitter->propagateToZ(&vtSeed, zFirstVTCluster);
-    }
-
     if (!mTrackFitter->fitTrackPointsOutward(seed, &vtSeed)) {
```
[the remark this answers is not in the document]
shahor02: The propagation to the 1st cluster to update is done internally in the TrackFitter.

rec/src/NA6PTrack.cxx
```diff
@@ -340,23 +82,16 @@ void NA6PTrack::addCluster(const ClusterType* clu, int cluIndex, double chi2)
...
-                                 double en, double mass, int charge)
+bool NA6PTrack::updateTrack(const NA6PBaseCluster& cl, float maxChi2)
 {
-  // RS: note: we assume p=e here to avoid problem with e+ e- interpreted as muon
-  double p = en * en - mass * mass;
-  if (p <= 0) {
-    printf("Anomalous kinematics: E:%e M:%e", en, mass);
-    exit(1);
+  auto chi2 = getPredictedChi2(cl);
+  if (chi2 > maxChi2 || update(cl)) {
```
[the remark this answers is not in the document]
shahor02: Fixed

rec/src/NA6PMatching.cxx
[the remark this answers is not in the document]
shahor02: Could you actually check that this is done?

fprino:
Hi Ruben,

thank you for your work on the native parameterization.

Concerning your question on the fitter and the propagation. 
The treatment of the propagation to the vertex in the CA tracker (and consequently in the FastFitter class) was implemented in different steps, and therefore it may need some refactoring now that we introduce the native parameterizaion. The first "add on" to the fitter was the option to control whether or not to propagate a track to the vertex. This was dictated by two reasons: 1) in the fit of the clusters of a cell the track should not be propagated to the vertex, 2) the muon spectrometer tracks should not be propagated to the vertex by default, but passed to the matching code. Afterwards, when this feature was already in place we added the feature of the vertex constrain, which was mainly aimed at testing the performance of MS only tracks. This has some interplay with the option to propagate or not to the primary vertex.
We relied in these implementation on the NA6PTrack, which was known to evolve once the native parameterization would have replaced the ExtTrackPar one, so I think that your proposal of working only with the parameterization in the fwd/bkwd propagation and avoid cloning the additional information (flags, indices, etc.) makes a lot of sense.

Of course, I fully trust you, so if you want to go ahead and implement the changes, this is perfectly fine.
If you think that a zoom chat is useful, I am available.

Cheers, Francesco

rec/include/NA6PTrack.h
```diff
@@ -140,123 +91,38 @@ class NA6PTrack
...
  protected:
-  double mMass = 0.140;                      // particle mass
-  double mMatchChi2 = 0.f;                   // total chi2
-  double mChi2VT = 0.f;                      // total chi2 VT
-  double mChi2MS = 0.f;                      // total chi2 MS
-  double mChi2VTOuter = 0.f;                 // total chi2 VT outward fit
-  double mChi2MSOuter = 0.f;                 // total chi2 MS outward fit
-  double mChi2VTRefit = 0.f;                 // total chi2 VT inward refit
-  double mChi2MSRefit = 0.f;                 // total chi2 MS inward refit
+  NA6PTrackParCov mExtTrack;                 // track params
```
fprino: Do we still need this data member mExtTrack?

galocco:
Hi Ruben,
as Francesco said, feel free to proceed. You definitely know this better than I do.
Cheers,
Giacomo

rec/include/NA6PTrack.h
shahor02: Indeed, forgot to remov it, fixed.

shahor02:
@fprino @galocco OK, thanks, then I'll proceed and we can discuss in zoom in case I see some problems.

fprino:
Hi Ruben,

I started in parallel to have a look at the HelixHelpers methods in o2 in order to port them to NA6PRoot in view of implementing the DCA fitter:
https://github.com/fprino/NA6PRoot/blob/dcafitter/rec/include/NA6PHelixHelpers.h

In this first implementation I had adapted the functions to those of NA60TrackPar.h of the old recNative directory. 
I checked them this morning against the new implementation of NA60TrackPar.h and the method getCircleParams seems to be missing, while it was present in the recNative implementation (and I was using it in the helix helpers). 
I guess it can be implemented in this way in the new class:
template <typename T>
std::array<T, 3> NA6PTrackPar::getCircleParams(float by) const
{
  // get circle params as {R, Xcenter, Zcenter}, for straight line just set to current coordinates
  constexpr T ZeroCurv = 1e-8;
  auto crv = getCurvature(by);
  if (std::abs(crv) > ZeroCurv) {
    auto crvi = T(1) / crv;
    T pxz2pz2 = T(1) + T(getTx()) * T(getTx());
    T cs =  T(1) / std::sqrt(pxz2pz2);
    T sn = getTx() * cs;
    return {std::abs(crvi), getX() + cs * crvi, getZ() - sn * crvi};
  } else {
    return {T(0), T(getX()), T(getZ())};
  }
}
Alternatively, I can also add this functionality in the helix helpers.
All the other functions used in HelixHelpers should work smoothly with the new NA6PTrackPar class.
Thank you!

shahor02:
Hi @fprino 
I think it should be simply:
```
template <typename T=float>
std::array<T, 3> NA6PTrackPar::getCircleParams(float by) const
{
  // get circle params as {R, Xcenter, Zcenter}, for straight line just set to current coordinates
  constexpr T ZeroCurv = 1e-8;
  auto crv = getCurvature(by);
  if (std::abs(crv) > kSmallKappa) {
    auto crvi = T(1) / crv;
     T sn = getTx(), cs = std::sqrt((T(1) - sn) * (T(1) + sn));
     return {std::abs(crvi), getX() + cs * crvi, getZ() - sn * crvi};
  } else {
     return {T(0), T(getX()), T(getZ())};
  }
}
```
If you want, I can add it to NA6PTrackPar. And, since now the NA6PTrackPar is a simplified version of the O2 TrackPar (no Alpha), with O2 XYZ changed to ZXY. So, the HelixHelper and DCAFitter needs just axes redifinition.

fprino:
Hi Ruben,

many thanks for adding getCircleParams.

In case it is useful to spare you some work, I have a version of HelixHelper here:
https://github.com/fprino/NA6PRoot/blob/dcafitter/rec/include/NA6PHelixHelpers.h
which should include all the functionalities needed by the DCAFitter.
In particular, for linesCrossInfo and circleLineCrossInfo I used the methods of the NA6PLine class.

Cheers, Francesco

shahor02:
Hi @fprino 
Thanks! You were faster. But I see a few places where the axes change needs to be corrected, could open a PR from your branch, it would be easier to comment.

I see you have substituted some code by NA6PLine methods, which obviously makes sense. I've just pushed to this branch NA6PLine constructor directly from the NA6PTrackPar, to avoid going via extracted momenta. Also changed it internals to `array<float,3>`. We could also change all the `float p[3]` arguments of the methods by `array<float,3>&`, which would allow us to shorten the code significantly.

fprino:
Hi Ruben,

I agree to update NA6PLine to update the arguments of NA6PLine methods to use a std::array instead of the native c++ arrays. I can take care if you are not yet on it.

I can open a PR for the helix helper and the DCA fitter from my branch, sure.
Only doubt about the git usage: in order to test the compilation with the main/origin branch I have modified NA6PTrack in order to add the methods getCircleParams and getCurvature that were missing in the original implementation (the one using ExtTrackPar). These changes in NA6PTrack do not need to be merged because they are overcome by your updates in NA6PTrack and NA6PTrackPar.
Should I drop them from the branch?

Cheers, Francesco

shahor02:
Hi @fprino ,
OK, I'll change the NA6PLine in my PR. For the NA6PHelixHelper, I meant to open a PR just to facilitated commenting the lines where I think the axis swap needs to be fixed. But indeed, it would be faster if I simply grab your file and fix it miself, it will anyway need to be tailored to the updated track model.

Cheers,
 Ruben

shahor02:
@galocco Yesterday we discussed with @fprino to simplify little bit the matching part by letting the matched track to have its refitted kinematics + chi2 and simply referring on the standalone VT and MS tracks. Is this OK for you?
I will also provide a reader to load the 3 streams together.

fprino:
Yes, please, proceed by grabbing the file from my fork.

> Hi @fprino , OK, I'll change the NA6PLine in my PR. For the NA6PHelixHelper, I meant to open a PR just to facilitated commenting the lines where I think the axis swap needs to be fixed. But indeed, it would be faster if I simply grab your file and fix it miself, it will anyway need to be tailored to the updated track model.
> 
> Cheers, Ruben

galocco:
Yes, that’s fine with me. Just to clarify, is the chi2 you’re referring to the matching chi2 or the fit chi2 of the VT+MS track?
> @galocco Yesterday we discussed with @fprino to simplify little bit the matching part by letting the matched track to have its refitted kinematics + chi2 and simply referring on the standalone VT and MS tracks. Is this OK for you? I will also provide a reader to load the 3 streams together.

shahor02:
@galocco I would store in the matched track both the matching and global refit chi2, w/o splitting the latter to VT and MS parts. I am not sure what is the meaning of such a splitting: if e.g. you are doing refit towards the vertex, you start in the MS with 5 DOFs, by the time you are starting to fit in the VT, the track is already well constrained. 
There is no simple way to define the reduced chi2 for the VT in this case, only the total full and reduced chi2's have more or less clear probabilistic meaning.

shahor02:
@fprino, for some reason, I cannot open your comments in the github browser...
Actually, the fixes were suggested by the chatgpt codex :). 
For the 1st fix, I think the problem is there also for the O2, I'll write to Max and Felix.
I've changed the NA6PVertexerTracklets 2nd fix to your version with some extra helpers added to NA6PLine. Also, for consistency, renamed the CrossInfo class in the NA6PHelixHelper to NA6PHelixHelper.

fprino:
Thank you Ruben for taking care of signalling the O2 issue to Max and Felix.
All the new updates in NA6PLine and NA6PVertexerTracklets look good to me.

For info: I have essentially completed the porting of DCAFitterN from o2 to NA6PRoot:
https://github.com/fprino/NA6PRoot/blob/dcafitter/rec/include/NA6PDCAFitter.h
There are some places (marked as TODO) which will need some updates to match the new NA6PTrackPar and NA6PTrackParCov features, in particular those related to track propagation and parent track covariance.
I can take care of them when this PR is merged, or downloading locally the pr_natrs1 branch

fprino:
Hi Ruben,

when doing some tests of the DCA fitter for secondary vertices I spotted a missing include of NA6PLine in the helix helper class (see https://github.com/shahor02/NA6PRoot/commit/c40d87bfc5279ca800204a101e3529190ed7f7cd).  This was not breaking the compilation, but gave an error at runtime.
I signal this issue in case you want to include this small modification of the helix helper class in your PR.

Thanks!

Cheers, Francesco

shahor02:
[perfrep_oldnew_psiBg1.pdf](https://github.com/user-attachments/files/29699481/perfrep_oldnew_psiBg1.pdf)
[perfrep_oldnew_psiBg0.pdf](https://github.com/user-attachments/files/29699492/perfrep_oldnew_psiBg0.pdf)

Comparison of performance with old/new code (the cuts in the new one need some tuning) from the na6prec output running over 10k J/psi with ( perfrep_oldnew_psiBg1.pdf) and w/o (perfrep_oldnew_psiBg1.pdf) hadronic bg. events generated with `genDimuonBgEvent.C`.
Plots are normalized to MC tracks reconstructable in the respective detectors (all layers contributing).

rec/src/NA6PTrackerCA.cxx
```diff
@@ -770,87 +809,99 @@ void NA6PTrackerCA::fitAndSelectTracks(const std::vector<TrackCandidate>& trackC
...
+      fitTrackFast.getOuterParam().invalidate();
     }
-    if (mPropagateTracksToPrimaryVertex)
-      mTrackFitter->propagateToZ(&fitTrackFast, mPrimVertPos[2]);
-    if (mDoTrackConstrainedToPrimVert && primVert) {
-      NA6PTrack constrainedTr = fitTrackFast;
-      bool vcOk = mTrackFitter->constrainTrackToVertex(&constrainedTr, *primVert);
-      if (vcOk) {
-        fitTrackFast.setVertexConstrainedParam(constrainedTr.getTrackExtParam());
-        fitTrackFast.setStatusConstrained(true);
```
galocco, line 828: Just a minor comment. I guess fitTrackFast.setStatusConstrained(true); should be kept or moved inside setVertexConstrainedParam.

shahor02: Isn't it what is done in https://github.com/shahor02/NA6PRoot/pull/99/changes/BASE..1860975e8b792c9d33a6b95710eb6083c4574e98#diff-4851f470db97690794150f6e2e78baa524fbec94a9bb2ee7ced962036bab0eb2R874-R878
then 
https://github.com/shahor02/NA6PRoot/blob/1860975e8b792c9d33a6b95710eb6083c4574e98/rec/src/NA6PFastTrackFitter.cxx#L43-L64 ?

galocco: Ah ok, so if I understand correctly, we don’t need `mStatusConstrained` anymore:
https://github.com/shahor02/NA6PRoot/blob/1860975e8b792c9d33a6b95710eb6083c4574e98/rec/include/NA6PTrack.h#L97

We were using this flag here:
https://github.com/shahor02/NA6PRoot/blob/1860975e8b792c9d33a6b95710eb6083c4574e98/test/runMSTrackMIDTrackletMatching.C#L362

and in other non-committed analysis macros.

shahor02: Thanks for spotting. Indeed, I did not use these flags but did not remove them.
Now the corresponding getters will automatically return false unless a successful fit was performed; see the last commit.

PR #101, opened 2026-04-17
First implementation of a class to manage VT pixel plane segmentation

fprino:
Hi,
with this PR, I ported the isInAcc method of KMCPixelPlane of the fast simulation into a new class NA6PVerTelSegmentation that in the next future can also manage the conversion of hit coordinate into indices of fired pixels, acting as an helper in the digitization of the VT.
For the time being, the method isInAcc can be used in the fast conversion of hits to clusters in order to discard the hits in the inactive regions of the VT plane.

The usage is simple, having a hit with coordinates xhit and yhit:
  NA6PVerTelSegmentation vt;
  int accOk = vt.isInAcc(xhit, yhit);
  if (accOk == 1) {
    // this is a good hit, to be accepted

Here are some plots showing the results for random generation of hits over one plane :
VertTelPlane

VertTelPlaneZoom

As compared to the implementation in the fast simulation, I added a selection to tag as out of acceptance the hits in the zone between the top half sensor unit and the bottom half sensor unit:
  if (y < DeadYBottom || y > DYSens - DeadYTop) return -1;
  int topbot = int(y / (DYSens / 2));
  y -= topbot * (DYSens / 2);
  if (y < DeadTopBotHalves / 2) return -1;
Let me know if this is correct.

Once this method is validated, I will add a call in the conversions of hits to recpoints.

### DIFF
--- a/sim/include/NA6PVerTelSegmentation.h
+++ b/sim/include/NA6PVerTelSegmentation.h
@@ -1,0 +1,58 @@
+// NA6PCCopyright
+
+// Based on:
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef NA6P_VERTEL_SEGMENTATION_H
+#define NA6P_VERTEL_SEGMENTATION_H
+
+#include <Rtypes.h>
+
+// define dead areas and segmentation of the VT modules
+
+class NA6PVerTelSegmentation
+{
+ public:
+  
+  NA6PVerTelSegmentation() = default;
+
+  void setOffsetX(float val) { mOffsX = val;}
+  void setOffsetY(float val) { mOffsY = val;}
+  void setInterChipGap(float val) { mInterChipGap = val;}
+  
+  int isInAcc(float x, float y) const;
+  
+ protected:
+  static const float XSizeTot;
+  static const float YSizeTot; // readout side
+  static const float DeadXLong;  // right end cap
+  static const float DeadXShort; // left end cap (readout)
+  static const float DeadYBottom; // bottom dead zone
+  static const float DeadYTop; // top dead zone
+  static const float DeadTopBotHalves; // dead space between top and bottom halves
+  static const float DeadXTile;  // dead space between tiles in X
+  static const float DeadXDataBackbone; // dead space between every segment (triplet of tiles)
+  static const int NXTiles;
+  static const int NXSegments;
+  static const int NYSensors;
+  static const float DXTile;
+  static const float DXSegment;
+  static const float DYSens;
+  
+  float mOffsX = 0.3;
+  float mOffsY = -0.3;
+  float mInterChipGap = 0.02;
+  
+  ClassDefNV(NA6PVerTelSegmentation, 1);
+};
+
+#endif
--- a/sim/include/NA6PVerTelSegmentation.h
+++ b/sim/include/NA6PVerTelSegmentation.h
@@ -22,34 +22,40 @@
 class NA6PVerTelSegmentation
 {
  public:
+  static constexpr float DeadXLong = 0.45;                            // right end cap
+  static constexpr float DeadXShort = 0.15;                           // left end cap (readout)
+  static constexpr float DeadYBottom = 0.0525;                        // bottom dead zone
+  static constexpr float DeadYTop = 0.0525;                           // top dead zone
+  static constexpr float DeadTopBotHalves = 0.012;                    // dead space between top and bottom halves
+  static constexpr float DeadXTile = 0.002;                           // dead space between tiles in X
+  static constexpr float DeadXDataBackbone = 0.006;                   // dead space between every segment (triplet of tiles)
+  static constexpr float XSizeTot = 12.9996 + DeadXShort + DeadXLong; // 13.5996;
+  static constexpr float YSizeTot = 13.5898 + DeadYBottom + DeadYTop; // 13.6948; // readout side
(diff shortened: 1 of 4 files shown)


sim/include/NA6PVerTelSegmentation.h
```diff
@@ -0,0 +1,57 @@
...
+  static const float DeadYTop;          // top dead zone
+  static const float DeadTopBotHalves;  // dead space between top and bottom halves
+  static const float DeadXTile;         // dead space between tiles in X
+  static const float DeadXDataBackbone; // dead space between every segment (triplet of tiles)
+  static const int NXTiles;
+  static const int NXSegments;
+  static const int NYSensors;
+  static const float DXTile;
+  static const float DXSegment;
+  static const float DYSens;
```
shahor02: These could be actually `static constexpr float XXX = value;`, all defined in the header. In the na60fastsim I was constrained by the old c++ standards.

shahor02:
Hi @fprino , looks correct! Eventually, we should translate the hits to chips' local frame and the segmentation should work directly with local coordinates, but at the moment this will work. 
Concerning the rejection on chip edges: this will be somewhat pessimistic as the overlaps between the staggered half-planes are supposed to cover the gaps: https://docs.google.com/presentation/d/15MOSR4kp8qqtr9uzT7IHPRipYIO8hK6Y4cXvR7SeOM4/edit?pli=1&slide=id.g33f99011aab_0_6#slide=id.g33f99011aab_0_6
At the moment, we can probably set the DeadYBottom to 0 and add its value to the DeadYTop (effectively shifting the active matrix to align it with the chip edge), and do the same with DeadXLong, to emulate the no-gaps situation.
Once we implement real staggered geometry, then we will account for the dead zones (and possibly 2 hits per plane).

fprino:
Hi Ruben,

thanks for the feedback. I moved the constants to the header as you suggested.

I also implemented an option to set the dead zones in the "inward" sides of the sensors to 0, to mimic the no-gap configuration. Note that I removed the line y = YSizeTot - y from the case x > 0. In this way, the y axis is directed "from the beam towards the outside" for all sensors, which seems to me the easiest choice to place the dead zones at the outer side of the sensor along y.
Does this sound correct to you?

Cheers, Francesco

shahor02:
Hi Francesco,

Thanks! I think it is ok to redefine the orientation of the sensors, we will need to address this again when we implement real staggered planes.
Shall I already merge this?

Cheers,
 Ruben

shahor02:
@fprino actually, I am wondering should not we put at once here https://github.com/shahor02/NA6PRoot/blob/7eed6321785eca2c223acdfab49ae5f250ce1e14/sim/src/NA6PVerTel.cxx#L626 the check 
```
static NA6PVerTelSegmentation vt;
if (vt.isInAcc(positionStop.X(), positionStop.Y())!=1) {
  return false;
}
```
?

fprino:
Yes, we can do it there. Do you think it is better to call this method at the level of hit creation, rather than at the level of hits-to-clusters?

shahor02:
Well, the effect is the same, if the hits-to-clusters is the equivalent of digitisation/clusterisation, then putting it there is more logical.

fprino:
Hi Ruben,

below you can see the modification in NA6PVerTelReconstruction to apply the dead zones at the level of hits-to-clusters.
I did not add it to this PR because it touches a file that you are modifying in PR #99 and, since my changes are small, it will be easy to add them once PR #99 is merged.

Cheers, Francesco

```
diff --git a/rec/include/NA6PVerTelReconstruction.h b/rec/include/NA6PVerTelReconstruction.h
index 16f4725..ddab87d 100644
--- a/rec/include/NA6PVerTelReconstruction.h
+++ b/rec/include/NA6PVerTelReconstruction.h
@@ -48,6 +48,7 @@ class NA6PVerTelReconstruction : public NA6PReconstruction
   void writeClusters() override;
   void closeClustersOutput() override;
   // fast method to smear the hits bypassing digitization and cluster finder
+  void setEmulateNoGaps(bool flag) { mEmulateNoGaps = flag; }
   void setClusterSpaceResolution(double clures) { mCluRes = clures; }
   void hitsToRecPoints(const std::vector<NA6PVerTelHit>& hits);
   NA6PTrackerCA* getTracker() const { return mVTTracker; }
@@ -71,6 +72,7 @@ class NA6PVerTelReconstruction : public NA6PReconstruction
   std::vector<NA6PVerTelCluster> mClusters, *hClusPtr = &mClusters; // vector of clusters
   TFile* mClusFile = nullptr;                                       // file with clusters
   TTree* mClusTree = nullptr;                                       // tree of clusters
+  bool mEmulateNoGaps = true;                                       // flag for hits in dead zones of sensors
   double mCluRes = 5.e-4;                                           // cluster resolution, cm (for fast simu)
   std::vector<NA6PVertex> mVertices, *hVerticesPtr = &mVertices;    // vector of vertices
   TFile* mVertexFile = nullptr;                                     // file with vertices
diff --git a/rec/src/NA6PVerTelReconstruction.cxx b/rec/src/NA6PVerTelReconstruction.cxx
index bfa534c..018006b 100644
--- a/rec/src/NA6PVerTelReconstruction.cxx
+++ b/rec/src/NA6PVerTelReconstruction.cxx
@@ -5,6 +5,7 @@
 #include <TRandom3.h>
 #include <fairlogger/Logger.h>
 #include "NA6PVerTelHit.h"
+#include "NA6PVerTelSegmentation.h"
 #include "NA6PTrackerCA.h"
 #include "NA6PVertexerTracklets.h"
 #include "NA6PVerTelReconstruction.h"
@@ -104,11 +105,16 @@ void NA6PVerTelReconstruction::setClusters(std::vector<NA6PVerTelCluster>& clust
 void NA6PVerTelReconstruction::hitsToRecPoints(const std::vector<NA6PVerTelHit>& hits)
 {
   int nHits = hits.size();
+  NA6PVerTelSegmentation vt;
+  if (mEmulateNoGaps)
+    vt.setStaggered(true);
   for (int jHit = 0; jHit < nHits; ++jHit) {
     const auto& hit = hits[jHit];
     double x = hit.getX();
     double y = hit.getY();
     double z = hit.getZ();
+    if (vt.isInAcc(x, y) != 1)
+      continue;
     double ex2clu = 5.e-4;
     double ey2clu = 5.e-4;
     if (mCluRes > 0) {

```

shahor02 (approved):
OK, thanks! I'll slightly modify it later to add it as configurableParam option.

PR #102, opened 2026-04-22
adding air layer between absorber and plug

scompar:
This is a modification of the absorber structure, allowing a layer of air between absorber and plug, to accommodate mechanical imprecision in the coupling of the two elements. It was suggested to perform these checks by the BE-EA engineers (see slides in https://indico.cern.ch/event/1673115/ for preliminary results).
Also, a new material corresponding to sintered graphite (density =1.9g/cm2) was added, as this material is apparently easier to work for creating a structure to insert the plug  (NA6PAbsorber).
This modification implies adding in the na6pLayout.ini file a series of new cards radAir[x]=y.y to specify, layer by layer, the thickness of this remaining air region.

Due to the increase in the number of volumes I had to increase in NA6PModule the following value --> static constexpr int MaxNonSensID = 300

Finally, also NA6PLayoutParam was modified to give default values to the radAir data cards.


sim/src/NA6PAbsorber.cxx
```diff
@@ -92,15 +97,74 @@ void NA6PAbsorber::createGeometry(TGeoVolume* world)
...
+  airVol->SetLineColor(kBlue - 10);
+
+  world->AddNode(
+    airVol,
+    composeNonSensorVolID(isl + 100),
+    new TGeoTranslation(0.0, 0.0, zpos)
+  );
+
+  // --- 3. PLUG (inside air)
+  if (!param.medAbsorberPlug[isl].empty() && param.radPlug[isl] > 0) {
```
shahor02: Hi @scompar ,
This is now within the `if (param.radAir[isl] > 0) {` scope, i.e. w/o air volume also the plug will not be added. Is this what we want? 
Also, could you please fix the formatting by running e.g. 
```
~/NA6PRoot$ git clang-format HEAD~3
```
then committing the fixed code?

scompar: Hello @shahor02 , I modified the code in such a way that the air layer is added only if the radAir[] is present and not zero. Also re-formatted following your suggestion

PR #103, opened 2026-06-09
Vertex Telescope digitization 

fprino:
This is a PR for the digitization of the Vertex Telescope.
In the first commit, only some basic features are implemented. Others will be added in the next days, but I thought it was useful to start discussing some aspects before moving ahead too much.

In NA6PVerTelSegmentation I added the methods needed to convert local coordinates on the sensor into a set of indices that identify uniquely a pixel in a sensor plane (up to 6x7 RSU).
I chose to use 4 indices: an rsu index (ranging from 0 to 41), a tile index (ranging from 0 to 11), row and column in each tile. These 4 indices identify uniquely a pixel on a sensor.

NA6PVerTelPreDigitContainer.h: this new class defines the "predigit" (or "summable digit") object, which is a float value per pixel that will store the deposited energy (or number of electrons in the pixel). The identification of a pixel is based on the 4 indices described above (rsu, tile, row, col).
Note, this is an intermediate format in the digitization process: the digits will be a list of fired pixels and will be obtained from the predigits in the finalization step of the digitization by applying the thresholds on the pixel energies.

NA6PVerTelDigitizer: this new class will steer the digitization step by running over the hits and converting them into digits. Currently the method processHit is just sketched and just reaches the point of converting the coordinates of the hit into the local coordinates on the sensor. However, before continuing with the implementation, I thought it was better to have a checkpoint on the structure I chose for the predigit (pixel indices) and for the segmentation.

@shahor02, @scompar, @galocco: any feedback is welcome!

### DIFF
--- a/base/include/NA6PGeometryManager.h
+++ b/base/include/NA6PGeometryManager.h
@@ -1,0 +1,54 @@
+// NA6PCCopyright
+
+// Based on:
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef NA6P_GEOMETRY_MANAGER_H
+#define NA6P_GEOMETRY_MANAGER_H
+
+#include <Rtypes.h>
+#include <TGeoMatrix.h>
+
+class TFile;
+class TGeoVolume;
+
+// helper class to interface to the geometry (martices, sizes) of sensors
+
+class NA6PGeometryManager
+{
+ public:
+  static constexpr int kNVTModulesPerLayer = 4;
+
+  NA6PGeometryManager() = default;
+  ~NA6PGeometryManager() = default;
+
+  bool loadGeometry(const char* filename = "geometry.root", const char* geoname = "NA6P");
+  bool isGeometryLoaded() const { return mGeoLoaded; }
+
+  const TGeoHMatrix& getMatrix(int jMod) const { return mMatrices[jMod]; }
+  float getModuleHalfX(int jMod) const { return mModuleHalfX[jMod]; }
+  float getModuleHalfY(int jMod) const { return mModuleHalfY[jMod]; }
+  float getModuleFullX(int jMod) const { return mModuleHalfX[jMod] * 2.0f; }
+  float getModuleFullY(int jMod) const { return mModuleHalfY[jMod] * 2.0f; }
+
+ private:
+  bool fillModuleSize(int jMod, TGeoVolume* vol);
+
+ protected:
+  std::vector<TGeoHMatrix> mMatrices{}; ///< local-to-global transforms
+  std::vector<float> mModuleHalfX;      ///< Module half length along x
+  std::vector<float> mModuleHalfY;      ///< Module half length along y
+  bool mGeoLoaded{false};               ///< flag for successful load of geo
+  ClassDefNV(NA6PGeometryManager, 1);
+};
+
+#endif
--- a/sim/include/NA6PVerTelDigit.h
+++ b/sim/include/NA6PVerTelDigit.h
@@ -18,6 +18,10 @@
 #include <Rtypes.h>
 
 // object for VT digits
+// NOTE: VTPixID uses C++ bit fields which ROOT cannot stream correctly.
+// Do not access mPixID branches directly via TTree::Scan() or TTree::Draw().
+// Always use the getters (getRSU(), getTile(), getRow(), getCol())
+// to access pixel indices.
 
 struct VTPixID {
 
@@ -59,30 +63,31 @@ class NA6PVerTelDigit
 {
  public:
   NA6PVerTelDigit() = default;
-  NA6PVerTelDigit(UShort_t detID, UShort_t rsu, UShort_t tile,
-                  UShort_t row, UShort_t col, int pid = -1);
+  NA6PVerTelDigit(uint16_t detID, const VTPixID& id, int pid = -1);
+  NA6PVerTelDigit(uint16_t detID, uint32_t rsu, uint32_t tile,
+                  uint32_t row, uint32_t col, int pid = -1);
 
-  UShort_t getDetectorID() const { return mDetectorID; }
+  uint16_t getDetectorID() const { return mDetectorID; }
   const VTPixID& getPixID() const { return mPixID; }
-  UShort_t getRSU() const { return mPixID.rsu; }
-  UShort_t getTile() const { return mPixID.tile; }
-  UShort_t getRow() const { return mPixID.row; }
-  UShort_t getCol() const { return mPixID.col; }
+  uint32_t getRSU() const { return mPixID.rsu; }
+  uint32_t getTile() const { return mPixID.tile; }
+  uint32_t getRow() const { return mPixID.row; }
+  uint32_t getCol() const { return mPixID.col; }
   int getParticleID() const { return mParticleID; }
 
-  void setDetectorID(int id) { mDetectorID = id; }
+  void setDetectorID(uint16_t id) { mDetectorID = id; }
   void setPixID(const VTPixID& id) { mPixID = id; }
-  void setRSU(int id) { mPixID.rsu = id; }
-  void setTile(int id) { mPixID.tile = id; }
-  void setRow(int id) { mPixID.row = id; }
-  void setCol(int id) { mPixID.col = id; }
+  void setRSU(uint32_t id) { mPixID.rsu = id; }
+  void setTile(uint32_t id) { mPixID.tile = id; }
+  void setRow(uint32_t id) { mPixID.row = id; }
+  void setCol(uint32_t id) { mPixID.col = id; }
   void setParticleID(int id) { mParticleID = id; }
 
   void print() const;
   std::string asString() const;
 
  protected:
-  UShort_t mDetectorID = 0; // the detector/sensor id
+  uint16_t mDetectorID = 0; // the detector/sensor id
   VTPixID mPixID;           // pixel identifier
   int mParticleID = -1;     // Particle ID
 
--- a/sim/include/NA6PVerTelDigit.h
+++ b/sim/include/NA6PVerTelDigit.h
@@ -63,9 +63,9 @@ class NA6PVerTelDigit
 {
  public:
   NA6PVerTelDigit() = default;
-  NA6PVerTelDigit(uint16_t detID, const VTPixID& id, int pid = -1);
+  NA6PVerTelDigit(uint16_t detID, const VTPixID& id);
   NA6PVerTelDigit(uint16_t detID, uint32_t rsu, uint32_t tile,
-                  uint32_t row, uint32_t col, int pid = -1);
+                  uint32_t row, uint32_t col);
 
   uint16_t getDetectorID() const { return mDetectorID; }
   const VTPixID& getPixID() const { return mPixID; }
@@ -73,7 +73,6 @@ class NA6PVerTelDigit
   uint32_t getTile() const { return mPixID.tile; }
   uint32_t getRow() const { return mPixID.row; }
   uint32_t getCol() const { return mPixID.col; }
-  int getParticleID() const { return mParticleID; }
 
   void setDetectorID(uint16_t id) { mDetectorID = id; }
   void setPixID(const VTPixID& id) { mPixID = id; }
@@ -81,7 +80,6 @@ class NA6PVerTelDigit
   void setTile(uint32_t id) { mPixID.tile = id; }
   void setRow(uint32_t id) { mPixID.row = id; }
   void setCol(uint32_t id) { mPixID.col = id; }
-  void setParticleID(int id) { mParticleID = id; }
 
   void print() const;
   std::string asString() const;
@@ -89,7 +87,6 @@ class NA6PVerTelDigit
  protected:
   uint16_t mDetectorID = 0; // the detector/sensor id
   VTPixID mPixID;           // pixel identifier
-  int mParticleID = -1;     // Particle ID
 
   ClassDefNV(NA6PVerTelDigit, 1);
 };
--- a/sim/include/NA6PVerTelDigit.h
+++ b/sim/include/NA6PVerTelDigit.h
@@ -1,0 +1,92 @@
+// NA6PCCopyright
+
+// Based on:
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef NA6P_VERTEL_DIGIT_H
+#define NA6P_VERTEL_DIGIT_H
+
+#include <Rtypes.h>
+
+// object for VT digits
+
+struct VTPixID {
+
+  static constexpr int kColBits = 10; // 0-161,  max 1023
+  static constexpr int kRowBits = 10; // 0-459,  max 1023
+  static constexpr int kTileBits = 4; // 0-11,   max 15
+  static constexpr int kRsuBits = 6;  // 0-41,   max 63
+
+  static_assert(kColBits + kRowBits + kTileBits + kRsuBits == 30,
+                "VTPixID fields must fit in 32 bits");
+
+  uint32_t col : kColBits = 0;
+  uint32_t row : kRowBits = 0;
+  uint32_t tile : kTileBits = 0;
+  uint32_t rsu : kRsuBits = 0;
+
+  uint32_t pack() const
+  {
+    return uint32_t(col) |
+           (uint32_t(row) << kColBits) |
+           (uint32_t(tile) << (kColBits + kRowBits)) |
+           (uint32_t(rsu) << (kColBits + kRowBits + kTileBits));
+  }
+
+  static VTPixID unpack(uint32_t val)
+  {
+    VTPixID id;
+    id.col = val & ((1 << kColBits) - 1);
+    id.row = (val >> kColBits) & ((1 << kRowBits) - 1);
+    id.tile = (val >> (kColBits + kRowBits)) & ((1 << kTileBits) - 1);
+    id.rsu = (val >> (kColBits + kRowBits + kTileBits)) & ((1 << kRsuBits) - 1);
+    return id;
+  }
+
+  ClassDefNV(VTPixID, 1);
+};
+
+class NA6PVerTelDigit
+{
+ public:
+  NA6PVerTelDigit() = default;
+  NA6PVerTelDigit(UShort_t detID, UShort_t rsu, UShort_t tile,
+                  UShort_t row, UShort_t col, int pid = -1);
+
+  UShort_t getDetectorID() const { return mDetectorID; }
+  const VTPixID& getPixID() const { return mPixID; }
+  UShort_t getRSU() const { return mPixID.rsu; }
+  UShort_t getTile() const { return mPixID.tile; }
+  UShort_t getRow() const { return mPixID.row; }
+  UShort_t getCol() const { return mPixID.col; }
+  int getParticleID() const { return mParticleID; }
+
+  void setDetectorID(int id) { mDetectorID = id; }
+  void setPixID(const VTPixID& id) { mPixID = id; }
+  void setRSU(int id) { mPixID.rsu = id; }
+  void setTile(int id) { mPixID.tile = id; }
+  void setRow(int id) { mPixID.row = id; }
+  void setCol(int id) { mPixID.col = id; }
+  void setParticleID(int id) { mParticleID = id; }
+
+  void print() const;
+  std::string asString() const;
+
+ protected:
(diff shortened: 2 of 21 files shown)


fprino:
Here are some plots from the checks I did on the segmentation, which illustrate the way I defined the idices ofr RSU and tiles.

1) RSU index vs local x and y (note the plotted value is actually rsu +1 in order to have it in the range 1-42, instead of 0-41 for proper visualization in the plot)
Segmentation-RSU-Index

2) TILE index vs local x and y (note the plotted value is actually tile +1 in order to have it in the range 1-12, instead of 0-11 for proper visualization in the plot)
Segmentation-Tile-Index

3) ROW vs vs local x and y 
Segmentation-Row

4) COL vs vs local x and y 
Segmentation-Col

5) residual between the local x and y coordinates of the pixel center and the hit, for hits in the active regions of hte sensors:
Pixel-Hit-LocalCoord-Residuals

fprino:
I updated the global->local conversion by adding the proper swaps of sign for the 4 modules of each plane in order to have the (0,0) at the bottom left corner, with the short end cap along xlocal on the left.
This is done in the digitizer, but it could be eventually moved to the rotation matrices.
Here is the resulting plots of mapping global to local x and y for each layer:
GlobToLoc

I also added some minimal code in NA6PVerTelDigitizer::processHit to show how the filling of pre-digits would be done: it is a very rough implementation, just meant to provide an idea of the functionalities that we could use from the segmentation and the pre-digits.

sim/include/NA6PVerTelPreDigitContainer.h
```diff
@@ -0,0 +1,116 @@
...
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef NA6P_VERTEL_PREDIGITCONTAINER_H
+#define NA6P_VERTEL_PREDIGITCONTAINER_H
+
+#include <Rtypes.h>
+
+// container of pre-digits per module
+
```
shahor02, line 23: I would define the global index as  a separate struct packed to a single int, it could be used also in other classes, e.g.

```
struct VTPixID {
  static constexpr int kColBits = 10; // 0-161,  max 1023
  static constexpr int kRowBits = 10; // 0-459,  max 1023
  static constexpr int kTileBits = 4; // 0-11,   max 15
  static constexpr int kRsuBits = 6;  // 0-41,   max 63
  
  uint16_t rsu  : kRsuBits = 0;
  uint16_t tile : kTileBits = 0;
  uint16_t row  : kRowBits = 0;
  uint16_t col  : kColBits = 0;
  
  uint32_t pack() const
  {
    return uint32_t(col) |
      (uint32_t(row)  << (kColBits)) |
      (uint32_t(tile) << (kColBits+kRowBits)) |
      (uint32_t(rsu) << (kColBits+kRowBits+kTileBits));
  }
  ...

  ClassDefNV(IDX,1);
};

struct PreDigit {
  VTPixID pixID;
  float charge = 0.f;
  int particleID = -1;
  ClassDefNV(PreDigit,1);
}
```

sim/include/NA6PVerTelSegmentation.h
```diff
@@ -45,7 +53,15 @@ class NA6PVerTelSegmentation
   void setInterChipGap(float val) { mInterChipGap = val; }
   void setStaggered(bool val);
 
-  int isInAcc(float x, float y) const;
+  bool localToIndices(float xloc, float yloc, int& rsu, int& tile, int& row, int& col) const;
+  int isInAcc(float xglo, float yglo) const;
+  bool indicesToLocal(int rsu, int tile, int row, int col, float& xloc, float& yloc) const;
```
shahor02: better use the same types as for indices, i.e. UShort_t or uint16_t, to allow direct assignment.

sim/include/NA6PVerTelSegmentation.h
```diff
@@ -45,7 +53,15 @@ class NA6PVerTelSegmentation
...
+
+ private:
+  bool computePixelIndices(float xloc, float yloc,
+                           float deadYBottom, float deadYTop,
+                           int& rsu, int& tile,
+                           int& row, int& col) const;
```
shahor02: idem

sim/src/NA6PVerTelDigitizer.cxx
```diff
@@ -0,0 +1,157 @@
...
+#include <fairlogger/Logger.h>
+#include <TGeoNode.h>
+#include <TGeoBBox.h>
+#include <TGeoManager.h>
+
+#include "NA6PLayoutParam.h"
+#include "NA6PVerTelHit.h"
+#include "NA6PVerTelDigitizer.h"
+
+ClassImp(NA6PVerTelDigitizer)
```
shahor02: ClassImp is not required anymore

sim/src/NA6PVerTelDigitizer.cxx
```diff
@@ -0,0 +1,157 @@
...
+
+  std::vector<bool> isMatrixLoaded;
+  isMatrixLoaded.assign(mNumberOfModules, false);
+  mModuleHalfX.assign(mNumberOfModules, 0.);
+  mModuleHalfY.assign(mNumberOfModules, 0.);
+  TGeoIterator next(gGeoManager->GetTopVolume());
+  TGeoNode* node = nullptr;
+  while ((node = next())) {
+    TString name = node->GetName();
+    if (name.BeginsWith("PixelSensor_")) {
```
shahor02: The sensors are already declared AlignableEntries, I think it is safer to access their volumes and matrices via a helper class like https://github.com/AliceO2Group/AliceO2/blob/323df58cec80b20d16cbd9d78ccab3888004fe8a/Detectors/Base/src/GeometryManager.cxx#L193)

shahor02 (commented):
Hi @fprino 
very good, just a few comments below.

sim/include/NA6PVerTelPreDigitContainer.h
fprino: Thanks! I implemented this struct in the new NA6PVerTelDigit class, so that it can be used also for storing the digits in the output root file.

sim/include/NA6PVerTelSegmentation.h
fprino: Done in the new commit

sim/include/NA6PVerTelSegmentation.h
fprino: Also done in the new commit

sim/src/NA6PVerTelDigitizer.cxx
fprino: Removed

sim/src/NA6PVerTelDigitizer.cxx
fprino: I agree. Will do it in the next round of updates. I guess we could have a single geometry helper class dealing  with both VT and MS volumes. Do you agree?

shahor02: Yes, this is what I meant.

sim/include/NA6PVerTelDigit.h
```diff
@@ -0,0 +1,92 @@
...
+
+  static constexpr int kColBits = 10; // 0-161,  max 1023
+  static constexpr int kRowBits = 10; // 0-459,  max 1023
+  static constexpr int kTileBits = 4; // 0-11,   max 15
+  static constexpr int kRsuBits = 6;  // 0-41,   max 63
+
+  static_assert(kColBits + kRowBits + kTileBits + kRsuBits == 30,
+                "VTPixID fields must fit in 32 bits");
+
+  uint32_t col : kColBits = 0;
```
shahor02, line 36: You defined data members as uint32_t (rather than e.g. uint16_t or UShort), which is good since it will almost guarantee that the compiler will pack it to underlying 32 bit integer and will probably give the best performance, since the 32 int is the native type.
But then, better to define the corresponding arguments as of the same type, so that one can call e.g.
```
VTPixID id;
localToIndices(xloc, yloc, id.rsu, id.tile, id.row, id.col);
```
(that said, it is more economical to have `localToIndices(flost xloc, float yloc, VTPixID& id);`

sim/include/NA6PVerTelDigit.h
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  ClassDefNV(VTPixID, 1);
+};
+
+class NA6PVerTelDigit
+{
+ public:
+  NA6PVerTelDigit() = default;
+  NA6PVerTelDigit(UShort_t detID, UShort_t rsu, UShort_t tile,
```
shahor02: I meant here and below one can use directly uint32_t.

fprino: Thanks for the suggestion. I will change it. 

However, I have an issue with the VTPixID object, when I write the NA6PVerTelDigits in a root file. I tried to implement the same structure used for hits and clusters: namely a TTree whose branches are an std::vector of NA6PVerTelDigits. When I read back the digits from the root file, the VTPixID data member is missing, as if it was not written.
 ```
***********************************************************************************************
*    Row   * Instance * VerTel.mD * VerTel.mP * VerTel.mP * VerTel.mP * VerTel.mP * VerTel.mP *
***********************************************************************************************
*        0 *        0 *         0 *           *           *           *           *       301 *
*        0 *        1 *         0 *           *           *           *           *       374 *
*        0 *        2 *         0 *           *           *           *           *       159 *
*        0 *        3 *         0 *           *           *           *           *       248 *
*

```

I checked that the object in memory (before pushing it into the std::vector) is correctly filled.
Is this due to some limitations in ROOT i/o for treating a struct like VTPixID?

Comment (shahor02) [reply sim/include/NA6PVerTelDigit.h]: Hi @fprino ,

I don't see VTPixID in the linkdef, could you add it? If the problem persist, could you send me the logs of writing and reading?

Comment (fprino) [reply sim/include/NA6PVerTelDigit.h]: Hi Ruben,

indeed I forgot to add VTPixID in the linkdef. By adding it, things improved: some values are actually stored in the TTree. However, when I print them with TTree::Scan or plot them from the TBrowser using the VTPixID data members, I get very strange values:

```
***********************************************************************************************
*    Row   * Instance * VerTel.mD * VerTel.mP * VerTel.mP * VerTel.mP * VerTel.mP * VerTel.mP *
***********************************************************************************************
*        0 *        0 *         0 *     15370 * 754974780 *  19726336 *     77056 *       301 *
*        0 *        1 *         0 *     19475 * 1.979e+09 *  24510464 *     95744 *       374 *
*        0 *        2 *         0 *     20536 * 2.667e+09 *  10420224 *     40704 *       159 *
*        0 *        3 *         0 *     38955 * 4.160e+09 *  16252928 *     63488 *       248 *
*        0 *        4 *         0 *     47234 * 2.130e+09 *   8323072 *     32512 *       127 *
*        0 *        5 *         0 *     61460 * 150995184 *  17367040 *     67840 *       265 *

```

rsu1

If instead I plot them using the getter methods of the class, I obtain the correct plot:
rsu2

It seems an issue with packing the bits of the struct into the root output.

I also tried to use UShort_t instead of uint32_t for the members of VTPixID, but also in this case the values are not correct:
```
***********************************************************************************************
*    Row   * Instance * VerTel.mD * VerTel.mP * VerTel.mP * VerTel.mP * VerTel.mP * VerTel.mP *
***********************************************************************************************
*        0 *        0 *         0 *     33802 *        15 *         0 *         0 *       301 *
*        0 *        1 *         0 *     34835 *        19 *         0 *         0 *       374 *
*        0 *        2 *         0 *     21560 *        20 *         0 *         0 *       159 *

```

Comment (shahor02) [reply sim/include/NA6PVerTelDigit.h]: This makes sense: apparently, the tree drawing does not know how to extract the packed value. Using class getter delegates it to the compiled code, which has no such issue.
Do you see this as a problem? If so, we can drop packing and use 3 uint8_t and one uint16_t, but this will lead to sizrof of 6 bytes instead of 4.

Comment (fprino) [reply sim/include/NA6PVerTelDigit.h]: In my opinion the possibility of directly drawing from the tree would be nice and user-friendly feature, but I do not see it as a "must": I do not think that browsing the digits from the TTree will be a frequent operation after the first debug phase.
A compromised solution could be to keep the packed object only for the predigits, which stay only in memory and are never written to root file (unless we will do embedding at the digit level), and use the 3 uint8_t and one uint16_t as data members of the digit. But we pay a price in disk space.
So, all in all, since you have confirmed that only the tree drawing is affected, I would tend to keep the packed struct.

File: sim/include/NA6PVerTelSegmentation.h
Comment (galocco) [inline sim/include/NA6PVerTelSegmentation.h:55]: Shouldn't this match pixChipOffsX = 0.29f in NA6PVerTel.cxx?

File: sim/src/NA6PVerTelDigitizer.cxx [hunk 8]
```diff
@@ -0,0 +1,217 @@
...
+#include "NA6PVerTelHit.h"
+#include "NA6PVerTelDigitizer.h"
+
+void NA6PVerTelDigitizer::init(const char* filename, const char* geoname)
+{
+  const auto& param = NA6PLayoutParam::Instance();
+  mNumberOfModules = param.nVerTelPlanes * kNModulesPerLayer;
+  mModules.resize(mNumberOfModules);
+  mMatrices.resize(mNumberOfModules);
+  mThresholds.assign(mNumberOfModules, kDefaultThresholdEl);
```
galocco: Just as a note for future reference: in principle, with MOSAIX, the threshold could also be adjusted independently for each tile

sim/include/NA6PVerTelSegmentation.h
fprino: Thanks. Fixed.

sim/src/NA6PVerTelDigitizer.cxx
fprino: I redefined the vector size to have one threshold per tile

sim/src/NA6PVerTelDigitizer.cxx
fprino: For info: the implementation of the helper class to access the AlignableEntries depends on the fixes in https://github.com/shahor02/NA6PRoot/pull/104 and https://github.com/shahor02/NA6PRoot/pull/105.

fprino:
The last commit implements the usage of NA6PMCTruthContainer to store the MC labels for digits.
This allows to store more than one particle ID per digit in case more particles have crossed the same pixel.
Here is an example for 50 central Pb-Pb events (NA49-like):
MCLabels-Digits

fprino:
Hi @shahor02 , @galocco , @scompar 

With the last commits, it seems to me that all the features that we discussed in the last NA60+ meeting are implemented. The code is functional and ready for deeper tests.
I also committed a simple macro to show how to read the VT digits and the MC truth for the next studies. 
So, I'd say that we reached a good check point for reviewing the implementation and merging this PR.

I am working on the rotation of the local frames for the VT modules (also discussed in the NA60+ meeting), but I'd include it in a different PR in the next days since it affects different classes and it is not directly related to the digitization features.

Thank you for the feedback and suggestions you gave so far and those that will come.

base/include/NA6PGeometryManager.h
```diff
@@ -0,0 +1,54 @@
...
+  float getModuleFullX(int jMod) const { return mModuleHalfX[jMod] * 2.0f; }
+  float getModuleFullY(int jMod) const { return mModuleHalfY[jMod] * 2.0f; }
+
+ private:
+  bool fillModuleSize(int jMod, TGeoVolume* vol);
+
+ protected:
+  std::vector<TGeoHMatrix> mMatrices{}; ///< local-to-global transforms
+  std::vector<float> mModuleHalfX;      ///< Module half length along x
+  std::vector<float> mModuleHalfY;      ///< Module half length along y
```
shahor02, line 49: Is not this available from the segmentation as a constant? Why do we need it dynamically?
Also, if we really need it, it is worth indicating by name that this is for the VT.

sim/include/NA6PVerTelPreDigitContainer.h
```diff
@@ -0,0 +1,138 @@
...
+  PreDigit* findDigit(ULong64_t key);
+  void addDigit(ULong64_t key, UShort_t rs, UShort_t tl, UShort_t rw, UShort_t cl, float ch, const NA6PMCComposedLabel& lbl);
+
+  UShort_t getDetectorIndex() const { return mDetectorIndex; }
+  void setDetectorIndex(UShort_t ind) { mDetectorIndex = ind; }
+
+  bool isEmpty() const { return mPreDigits.empty(); }
+  static ULong64_t getOrderingKey(UShort_t rsu, UShort_t tile, UShort_t row, UShort_t col)
+  {
+    VTPixID id;
```
shahor02: This would be faster to construct in-place as 
```
VTPixID id{.rsu=rsu, .tile = tile, .row = row, .col = col};
```

sim/src/NA6PVerTelDigitizer.cxx
```diff
@@ -0,0 +1,185 @@
...
+                   return true;
+                 return detID2Layer(hits[idx].getDetectorID()) == layer;
+               })) {
+    processHit(hits[i]);
+  }
+  finalizeDigits();
+  writeDigits();
+}
+
+void NA6PVerTelDigitizer::getHitLocalCoord(NA6PVerTelHit hit, double xyzLocS[3], double xyzLocE[3])
```
shahor02: Pass by cons ref, `getHitLocalCoord(const NA6PVerTelHit& hit, ...)`

sim/src/NA6PVerTelDigitizer.cxx
```diff
@@ -0,0 +1,185 @@
...
+    // swap x and y
+    xyzLocS[0] = mGeoManager.getModuleFullX(modID) - xyzLocS[0];
+    xyzLocS[1] = mGeoManager.getModuleFullY(modID) - xyzLocS[1];
+    xyzLocE[0] = mGeoManager.getModuleFullX(modID) - xyzLocE[0];
+    xyzLocE[1] = mGeoManager.getModuleFullY(modID) - xyzLocE[1];
+  } else if (modID % NA6PGeometryManager::kNVTModulesPerLayer == 3) {
+    // swap y
+    xyzLocS[1] = mGeoManager.getModuleFullY(modID) - xyzLocS[1];
+    xyzLocE[1] = mGeoManager.getModuleFullY(modID) - xyzLocE[1];
+  }
```
shahor02, line 112: Instead of these swaps, shouldn't we rotate the sensors when placing them on the station?

sim/src/NA6PVerTelDigitizer.cxx
```diff
@@ -0,0 +1,185 @@
...
+}
+
+void NA6PVerTelDigitizer::processHit(NA6PVerTelHit hit)
```
shahor02: `...(const NA6PVerTelHit& hit)`

sim/src/NA6PVerTelSegmentation.cxx
```diff
@@ -18,49 +18,117 @@ void NA6PVerTelSegmentation::setStaggered(bool val)
...
+  int rsuX = segm / NSegmentsPerRSU;
+  int segInRsu = segm % NSegmentsPerRSU;
+  rsu = static_cast<UShort_t>(sens * (NXSegments / NSegmentsPerRSU) + rsuX);
+  tile = static_cast<UShort_t>(topbot * (NSegmentsPerRSU * NTilesPerSegment) + segInRsu * NTilesPerSegment + tileIdx);
+  col = static_cast<UShort_t>(xloc / ActiveDX * NColsPerTile);
+  row = static_cast<UShort_t>(yloc / ActiveDYTile * NRowsPerTile);
+  return true;
+}
+
+int NA6PVerTelSegmentation::isInAcc(float xglo, float yglo) const
```
shahor02: Once we are dealing with placement matrices, this check with the global coordinates is error-prone.
Shouldn't we simply convert to local coordinates like in the getHitLocalCoord ?

sim/src/NA6PVerTelDigitizer.cxx
fprino: Hi Ruben,

yes, sure, it is better to rotate the sensor, a we discussed in the NA60+ meeting. 
I already prepared the modifications in the geometry to rotate the sensors:
https://github.com/shahor02/NA6PRoot/commit/7fd6e25d6795b260024abec1791d97fe95bf72c2

The corresponding changes in the digitizer are straightforward and I already tested them locally.

I thought to commit them in a different PR, but if you prefer to have them already in this PR, I can add them.
Let me know.

Cheers, Francesco

base/include/NA6PGeometryManager.h
fprino: Yes, they can also be taken from the segmentation.
I thought that it was clearer (and safer) to use the same source (the geometry) as for the transformation matrices, since these lengths are used to shift the local reference frame of the module so as to center it in a corner of the sensor. If you prefer to take these numbers from the segmentation, I can change.

sim/src/NA6PVerTelDigitizer.cxx
shahor02: Ah, sorry, since you were already applying the matrices I thought you wanted to limit their usage for the alignment only. It is fine to have it in a separate PR.

sim/src/NA6PVerTelSegmentation.cxx
fprino: I am wondering if we still this method isInAcc.
This was the first method ported from the fast simulation in order to switch off the hits in the dead regions during the hits->cluster "by pass.
Adding here the conversion from local to global would require to add a dependence of the segmentation class on the geometry only for this purpose. 
The easiest solution is in my opinion to drop this method and modify NA6PVerTelReconstruction::hitsToRecPoints to compute the local coordinates of the hit and pass them to the segmentation.
I can do it, if you agree, but I am afraid it may interfere with your PR for the native track parameterization.
Let me know.

shahor02: Indeed, better to drop it, getHitLocalCoord should do the job.

fprino: I committed the modifications to use local coordinates in NA6PVerTelSegmentation::isInAcc and use getHitLocalCoord in the method doing hits to rec points.

Since isInAcc is used only in the hitsToRecPoints "bypass", I preserved the possibility to set an interChipGap and to emulate the 2x2 sensor geometry and overcome the acceptance gaps due to sensor edges via the setStaggered setting. On the long term, both these features can be implemented in the VT geometry. But for the time being, I thought it was good to keep these functionalities via the segmentation class.

PR #107, opened 2026-07-02
Fix for root streaming issue of VTPixID objects into TTree

fprino:
This modification fixes an issue of double free or corruption (!prev) which I spotted on TTrees of digits when doing larger number of events.
It seems it was due to the streaming of the VTPixID object.

### DIFF
--- a/sim/include/NA6PVerTelDigit.h
+++ b/sim/include/NA6PVerTelDigit.h
@@ -18,45 +18,39 @@
 #include <Rtypes.h>
 
 // object for VT digits
-// NOTE: VTPixID uses C++ bit fields which ROOT cannot stream correctly.
-// Do not access mPixID branches directly via TTree::Scan() or TTree::Draw().
-// Always use the getters (getRSU(), getTile(), getRow(), getCol())
-// to access pixel indices.
+// VTPixID stores rsu/tile/row/col packed into a single uint32_t,
+// to ensure correct ROOT streaming.
+// Always use the getters/setters to access pixel indices.
 
 struct VTPixID {
-
   static constexpr int kColBits = 10; // 0-155,  max 1023
   static constexpr int kRowBits = 10; // 0-443,  max 1023
   static constexpr int kTileBits = 4; // 0-11,   max 15
   static constexpr int kRsuBits = 6;  // 0-41,   max 63
-
   static_assert(kColBits + kRowBits + kTileBits + kRsuBits == 30,
                 "VTPixID fields must fit in 32 bits");
 
-  uint32_t col : kColBits = 0;
-  uint32_t row : kRowBits = 0;
-  uint32_t tile : kTileBits = 0;
-  uint32_t rsu : kRsuBits = 0;
+  uint32_t mPacked = 0; // the ONLY streamed member
 
-  uint32_t pack() const
-  {
-    return uint32_t(col) |
-           (uint32_t(row) << kColBits) |
-           (uint32_t(tile) << (kColBits + kRowBits)) |
-           (uint32_t(rsu) << (kColBits + kRowBits + kTileBits));
-  }
+  uint32_t getCol() const { return mPacked & ((1u << kColBits) - 1); }
+  uint32_t getRow() const { return (mPacked >> kColBits) & ((1u << kRowBits) - 1); }
+  uint32_t getTile() const { return (mPacked >> (kColBits + kRowBits)) & ((1u << kTileBits) - 1); }
+  uint32_t getRsu() const { return (mPacked >> (kColBits + kRowBits + kTileBits)) & ((1u << kRsuBits) - 1); }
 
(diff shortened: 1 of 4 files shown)


sim/include/NA6PVerTelDigit.h
```diff
@@ -18,45 +18,39 @@
...
+  uint32_t getTile() const { return (mPacked >> (kColBits + kRowBits)) & ((1u << kTileBits) - 1); }
+  uint32_t getRsu() const { return (mPacked >> (kColBits + kRowBits + kTileBits)) & ((1u << kRsuBits) - 1); }
+
+  void setCol(uint32_t v) { mPacked = (mPacked & ~(((1u << kColBits) - 1))) | (v & ((1u << kColBits) - 1)); }
+  void setRow(uint32_t v) { mPacked = (mPacked & ~(((1u << kRowBits) - 1) << kColBits)) | ((v & ((1u << kRowBits) - 1)) << kColBits); }
+  void setTile(uint32_t v) { mPacked = (mPacked & ~(((1u << kTileBits) - 1) << (kColBits + kRowBits))) | ((v & ((1u << kTileBits) - 1)) << (kColBits + kRowBits)); }
+  void setRsu(uint32_t v) { mPacked = (mPacked & ~(((1u << kRsuBits) - 1) << (kColBits + kRowBits + kTileBits))) | ((v & ((1u << kRsuBits) - 1)) << (kColBits + kRowBits + kTileBits)); }
 
+  uint32_t pack() const { return mPacked; }
   static VTPixID unpack(uint32_t val)
```
shahor02: Do we need this counter-intuitively named methods?

sim/include/NA6PVerTelPreDigitContainer.h
```diff
@@ -81,24 +81,28 @@ class NA6PVerTelPreDigitContainer
   bool isEmpty() const { return mPreDigits.empty(); }
   static ULong64_t getOrderingKey(UShort_t rsu, UShort_t tile, UShort_t row, UShort_t col)
```
shahor02, line 82: Since our key is 32 bit, we can change here and below ULong64_t to uin32_t (also the map of fired pixels in the NA6PVerTelPreDigitContainer to `std::map<uint32_t, PreDigit> mPreDigits;`

shahor02:
Was the  `double free or corruption` happening when writing the tree or at reading? Do you have a full log?

fprino:
Hi Ruben,
the error was occurring when reading. 
There was no additional information in the log, neither when using the plotting macro, nor when doing a basic test in root:

root [0] TFile* fd = TFile::Open("DigitsVerTel.root")
(TFile *) 0x1a94f030
root [1] TTree* td = (TTree*)fd->Get("digitsVerTel");
root [2] std::vector<NA6PVerTelDigit> vtDigits, *vtDigitsPtr = &vtDigits;
root [3] td->SetBranchAddress("VerTel", &vtDigitsPtr);
root [4] td->GetEvent(0)
(int) 108804
root [5] td->GetEvent(1)
(int) 105874
root [6] td->GetEvent(2)
double free or corruption (!prev)

shahor02:
Thanks, if you still have this file, could you please put it e.g. on the cernbox? I am curious what could be the reason of the failure.

sim/include/NA6PVerTelDigit.h
fprino: I have removed them in the 2nd commit. They were used in 2 methods of the pre-digit container but I adapted the code in those places.

sim/include/NA6PVerTelPreDigitContainer.h
fprino: The idea behind the 64 bits was to reserve some space for the possible need of using ROFs.

shahor02: ok

fprino:
Hi Ruben,

here is the directory on cernbox with the hits and the digits https://cernbox.cern.ch/s/F6WZsMNWiXQyCHu.
Note that this was a test using the updated geometry with rotations of the local reference system of the modules (from the branch https://github.com/fprino/NA6PRoot/tree/rotationVT), so the geometry.root should be taken from the same directory.
I run the digitazion with a modified NA6PVerTelDigitizer::getHitLocalCoord to match with this geometry (i.e. w/o the swaps of the coordinates).
The digitization run smoothly.
The issue appears when reading the root file of the digits.

shahor02:
Thanks! I've got the same error and the chatGPT/codex says:
```
• The current VTPixID stores col, row, tile, and rsu as C++ bit fields, and NA6PVerTelDigit persists that struct inside a std::vector. That is the weak point: ROOT’s generated streamer can build incorrect collection state for such a layout, which matches the crash during
  TGenCollectionProxy::Allocate and std::vector<NA6PVerTelDigit> growth.
```
and suggest modifying the NA6PVerTelDigit exactly like you did!

shahor02 (approved):

PR #111, opened 2026-07-08
Add realistic VT geometry

galocco:
This PR adds the latest version of the VT geometry designed by Silvia Coli.

The geometry is implemented from scratch. I'm not sure this is the optimal strategy, but it seems to be the approach used in ALICE. An alternative would be to convert the CAD geometry to GDML and import it directly; however, it is not clear to me why this approach is not used in ALICE, since implementing the geometry manually is prone to errors.

I tried to use LLMs to automate the implementation, since STEP files are text-based. However, at the moment they are not good at this task.

I also added a few parameters to `NA6PLayoutParams` to allow testing different sensor-staggering configurations, and removed duplicated geometry constants.

VT_front

VT_back


galocco:
@shahor02 @fprino I think this PR is ready to be reviewed

shahor02 (approved):
Thanks @galocco (and Silvia!)

I've checked the sim/rec, both seem to work fine, even though now the Z positions of clusters on the same layer are staggered by a few mm. We may need to add some adjustments for that.
The reason why Alice did not use CAD->TGeo is that till recent times it did not account for the materials.  Recently it was fixed by Sandro, afaik, CAD->TGeo will be used for the Alice3, don't know if it is 100% ready.

I would also wait for @fprino's approval before merging.

sim/include/NA6PVerTelSegmentation.h
```diff
@@ -70,14 +71,17 @@ class NA6PVerTelSegmentation
...
                            UShort_t& rsu, UShort_t& tile,
                            UShort_t& row, UShort_t& col) const;
 
  protected:
-  float mOffsX = 0.29;
-  float mOffsY = -0.31;
+  float mOffsX = 0.f;
+  float mOffsXFront = 0.f;
+  float mOffsXBack = 0.f;
+  float mOffsY = 0.f;
```
fprino, line 84: These parameters (mOffsX, mOffsY) were added in the segmentation to account for the dead regions for the case of hits->recpoints. Intead in the "standard" path of hits->digits followed by digits->recpoints they are not used because this is treated via the global->local transformation matrices of each sensor. We can maybe consider to drop these quantities in the segmentation and devise another way to account for the dead areas in the hits->recpoints case. This can go in a different PR, I just mentioned it here to keep track of this aspect

fprino:
Hi Giacomo,
looks very good! Many thanks!
I posted a minor comment related to the usage of the offsets in the segmentation class, which is something that we can modify in the future.
As far as I am concerned, it can be merged.
Cheers, Francesco

fprino (approved):

PR #115, opened 2026-07-09
add test/produceAOD.C

dsekihat:
add test/produceAOD.C
This macro produces analysis object data after simulation and reconstruction.
This is useful for studying physics performance.

### DIFF
--- a/test/produceAOD.C
+++ b/test/produceAOD.C
@@ -1,0 +1,388 @@
+#if !defined(__CINT__) || defined(__MAKECINT__)
+#include <vector>
+#include <cmath>
+
+#include <TFile.h>
+#include <TTree.h>
+#include <TParticle.h>
+
+#include "NA6PTrack.h"
+#include "NA6PVertex.h"
+#include "NA6PMatching.h"
+#endif
+
+// root -b -q produceAOD.C++g
+
+void produceAOD(
+const char* fileNameVerTel = "VerticesVerTel.root",
+const char* fileNameTracksVerTel = "TracksVerTel.root",
+const char* fileNameTracksMuonSpec = "TracksMuonSpec.root",
+const char* fileNameTracksMatching = "TracksMatching.root",
+const char* fileNameMC = "MCKine.root",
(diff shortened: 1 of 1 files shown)


shahor02:
Hi @dsekihat ,

Could you explain how such Alice AOD-like structure will be used w/o actually having an equivalent of the Alice analysis framework? I think in the absence of such a framework the object-wise storage in the trees referring to each other via indices would be more convenient, we just need a reader class which consistently loads different reco and MC parts of the same event. Shall we discuss this at the next meeting?

dsekihat:
Thank you for the good suggestion. I agree. I did not think about carefully, but think about only for-loop analysis. Let me close this PR, and come back with a hopefully better PR.

shahor02:
Sorry, I did not mean it should be closed, just wanted to understand how Allice AOD-like trees could be used. If this output is more convenient for your analysis, we can surely merge it (perhaps under a more specific name).

Just in case it could be useful: in my PR I've added test/runPerfStudy.C macro, which runs over the output of the na6prec (one can select what to check via the arguments checkVTTracks, checkMSTracks, checkMatches and checkDimu) and produces control trees and histos with MC and reco results. 
For the reconstructed dimuons you have the full like https://github.com/shahor02/NA6PRoot/blob/14f34bbafbe8609eb4fbfae89817a07d19529a79/test/runPerfStudy.C#L624-L634
E.g. to plot the differences between the reconstructed and generated dimuon mass vs generated rapidity for MC pairs with both tracks matchable (== reconstructible both with VT and MS with all layers contributing) and requiring that both muons are correctly reconstructed:
```
#after running
#na6prec
#root -b -q $NA6PINSTALLDIR/share/test/runPerfStudy.C+

TFile *_file0 = TFile::Open("recperf.root");
perfDimu->Draw("genDimu.M() - recDimuMatch.M() : genDimu.Rapidity()","matchable && recMatch0ID==genMu0ID && recMatch1ID==genMu1ID","colz");
```
image

PR #117, opened 2026-07-09
add a reader class

dsekihat:
Following the suggestion, I moved the event synchronization logic into NA6PEventReader. It consistently opens and loads VerticesVerTel, TracksVerTel, TracksMuonSpec, TracksMatching and MCKine for the same event entry.
An example macro is test/readEvent.C.

### DIFF
--- a/rec/include/NA6PEventReader.h
+++ b/rec/include/NA6PEventReader.h
@@ -1,0 +1,256 @@
+#ifndef NA6P_EVENT_READER_H
+#define NA6P_EVENT_READER_H
+
+#include "TFile.h"
+#include "TTree.h"
+#include "TParticle.h"
+
+#include "NA6PVertex.h"
+#include "NA6PTrack.h"
+#include "NA6PMatching.h"
+
+#include <cstdint>
+#include <memory>
+#include <vector>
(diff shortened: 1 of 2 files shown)


rec/include/NA6PEventReader.h
```diff
@@ -0,0 +1,256 @@
...
+  NA6PEventReader(
+    const char* fileNameVerTel         = "VerticesVerTel.root",
+    const char* fileNameTracksVerTel   = "TracksVerTel.root",
+    const char* fileNameTracksMuonSpec = "TracksMuonSpec.root",
+    const char* fileNameTracksMatching = "TracksMatching.root",
+    const char* fileNameMC             = "MCKine.root",
+    bool readMC = true
+  )
+  {
+    openFile(mFileVerTel, fileNameVerTel);
```
shahor02: Thanks! I would foresee a possibility to skip some trees, since the na6pre can be used for e,g, VT reconstruction only.

dsekihat: Hello Ruben, Thank you for your comments. It is included.

shahor02 (approved):

PR #125, opened 2026-07-15
Option for muon reco as track in MS matched to tracklet in MID

fprino:


rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -149,7 +163,194 @@ void NA6PMuonSpecReconstruction::closeTracksOutput()
...
+      zSecond = clu1.getZ();
+    }
+    if (!Propagator::Instance()->propagateToZ(outTr, zFirst, fitter->getPropOpt())) {
+      // track not refitted: save with specific status flag
+      tr.setStatusMS(NA6PTrack::kMSMatchedToMIDnotRefitted);
+      mTracks.push_back(tr);
+      continue;
+    }
+    bool success = false;
+    if (outTr.update(clu1)) {
```
shahor02: The order of adding clusters is irrelevant, now they are sorted internally in the NA6PFastTrackFitter, so these lines can go outside the `if`.
But you are sorting the clusters in Z, propagating to zFirst, then unconditionally updating with clu1, even if it corresponds to zSecond (hinted by the codex).

In principle, the NA6PFastTrackFitter::fitSeed now supports an incremental fit: you could simply add the MID clusters (only, w/o adding those of the MS part) and call `fitSeedOutward(outTr, false)` 
Then, if the full refit is needed, just add the MS clusters and call fitSeed... with covMat reset request (I've realized that the `if (mNClusters < 2)` check is not needed anymore in the `NA6PFastTrackFitter::fitSeed`

rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -149,7 +163,194 @@ void NA6PMuonSpecReconstruction::closeTracksOutput()
...
 }
+
+void NA6PMuonSpecReconstruction::runMSTrackMIDTrackletMatching()
+{
+  const auto& param = NA6PRecoParam::Instance();
+  auto& clusters = getClusters();
+
+  std::vector<NA6PTrack> trks = mMSTracker->getTracks();
+  int nTrks = trks.size();
+  std::vector<std::pair<NA6PMuonSpecCluster, NA6PMuonSpecCluster>> trkltsMID = mMSTracker->findTracklets(9, 10, clusters, mPrimaryVertex);
```
shahor02: In the MSTracker initialization you set the layers dynamically, `setStartLayer(param.vtNLayers)`, perhaps it is better to avoid hardcoding also here.

rec/include/NA6PTrack.h
```diff
@@ -89,13 +100,13 @@ class NA6PTrack : public NA6PTrackParCov
...
   std::array<int, kMaxLr> mClusterIndices{}; // cluster indices
   std::array<int, kMaxLr> mClusterPartID{};  // particle ID (per cluster) // RSTOD why this is needed? This info must be available from the cluster indices
-
   float mChi2 = 0.f;                         // total chi2
   float mChi2Outer = 0.f;                    // total chi2 outward fit
   uint32_t mClusterMap = 0;                  // pattern of clusters per layer
   int mNClusters = 0;                        // total hits
   int mParticleID = -1;                      // particle ID (MC truth)
   int mCAIteration = -1;                     //! CA iteration (for debug)
+  short mStatusMS = kNotMS;                  // status of MS track
```
shahor02, line 109: reset should also assign this status? That said, we should consider implementing separate 
VT and MS track classes and to template TrackerCA to account for that.

rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -74,6 +87,7 @@ void NA6PMuonSpecReconstruction::hitsToRecPoints(const std::vector<NA6PMuonSpecM
 {
   int nHits = hits.size();
   const auto& layout = NA6PLayoutParam::Instance();
+
```
shahor02, line 90: Another comment from the codex:
```
Cluster index fragility: internally produced clusters in hitsToRecPoints() are not assigned mClusterIndex, while the new matching path relies on getClusterIndex() for lookup after CA sorting. mClusterIndex is uninitialized in rec/include/NA6PBaseCluster.h:73. External setClusters() does assign indices, but owned mClusters do not.

The fragile part is this chain:

  1. NA6PTrack::addCluster() stores clu->getClusterIndex(), not the cluster’s current vector position.
     See rec/src/NA6PTrack.cxx:67.

  2. runMSTrackMIDTrackletMatching() later uses those stored values as “original IDs” to recover clusters after findTracklets() reorders the shared cluster vector.
     See rec/src/NA6PMuonSpecReconstruction.cxx:190.

  3. But cluster indices are only assigned in setClusters():
     rec/src/NA6PMuonSpecReconstruction.cxx:76

  For externally supplied clusters, this is okay because setClusters() does:

  for (size_t i = 0; i < clusters.size(); ++i) {
    clusters[i].setClusterIndex(static_cast<int>(i));
  }

  The issue is the internally owned path: hitsToRecPoints() creates clusters in mClusters, but does not call setClusterIndex() on them. So mClusterIndex stays whatever the default constructor leaves it as. In rec/include/
  NA6PBaseCluster.h:73, it is declared as:

  int mClusterIndex; //! Transient index for track-cluster association

  No initializer. That means owned clusters can carry undefined index values.
```

fprino: Indeed, I was aware of this fragility.

The vector mClusters is used in hitsToRecPoints to fill the clusters and is then saved into the output root file. 
In this phase, the mClusterIndex values are not needed.
In the NA6PRec workflow, clusters once computed (in hitsToRecPoints or in the clusterizer) are immediately saved into a root file. 
Then, for the next steps they are read from the root file and setClusters is called, which sets the cluster indices based on the position f the cluster in the TTree.
This works in the NA6PRec reco chain, but indeed it exposes to the risk of issues.

One idea to consolidate this could be to set the indices also in hitsToRecPoints (and equivalently in the clusterizer), so that if one uses the directly object mClusters instead of reading it from the root file, it will have the proper indices.
Another option could be to make the data member not transient, but this would increase the size of the clusters.
Do you have better ideas?

shahor02: In principle, I would get rid of the index completely: given the cluster pointer and its container, one can always get the index as e.g. `std::dist(&container.first(), clusPtr)`. 
I can implement this change once these PRs are merged.

rec/src/NA6PMuonSpecReconstruction.cxx
fprino: Done, thanks!

rec/include/NA6PTrack.h
fprino: I added the reset of the status for the time being.

rec/src/NA6PMuonSpecReconstruction.cxx
fprino: I tried to implement as you suggested: I first added the MID clusters to the fitter, called fitSeedOutward, the I added the 4 MS clusters and called fitSeedInward.
zFirst and zSecond are no longer needed.
The results did not change for the file with 5000 J/psi I use as a test.
Can you please check that I implemented properly what you suggested?

rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -149,7 +163,170 @@ void NA6PMuonSpecReconstruction::closeTracksOutput()
...
+    if (!Propagator::Instance()->propagateToZ(outTr, zToProp, fitter->getPropOpt())) {
+      // track not refitted: save with specific status flag
+      tr.setStatusMS(NA6PTrack::kMSMatchedToMIDnotRefitted);
+      mTracks.push_back(tr);
+      continue;
+    }
+    fitter->addCluster(clu1);
+    fitter->addCluster(clu2);
+    outTr.resetClusters();
+    if (!fitter->fitSeedOutward(outTr, false)) {
```
shahor02: Now it is ok except that `fitter->fitSeed...` returns a positive sum of  chi2s of all updates done in case of succes and -1 in case of failure (failed per-update check of `getPredictedChi2 < mMaxChi2Cl` is also treated as a failure, in this case, since you are extracting the fitter from the TrackerCA, where the mTrackFitter->setMaxChi2Cl(maxChi2TrClu) was set, you will inherit the mMaxChi2Cl from the general MS tracking settings.). So, it should be like
```
if (fitter->fitSeedOutward(outTr, false)<0) { // update by MID has failed
 ...
}
```

rec/src/NA6PMuonSpecReconstruction.cxx
fprino: I am not sure that this will work because the vector of clusters is sorted by the tracker.
The cluster index retrieved via track.getClusterIndex points instead to the "original" index of the clusters ordered as in the root file (i.e. before the sorting done by the tracker). 
This choice was made to allow retrieving the correct clusters from the root files for tracks and clusters after the reconstruction.

shahor02: I see, but I would also change this: this sorting of the vector, instead of using the sorted index, involves per-cluster copy of all elements. If for some reason the clusters cannot be prepared in advance in the needed order (eventually, there will also be a time dimension), it is cheaper to sort a vector<int> and leave the input clusters const.

rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -149,7 +163,170 @@ void NA6PMuonSpecReconstruction::closeTracksOutput()
...
   writeTracks();
 }
+
+void NA6PMuonSpecReconstruction::runMSTrackMIDTrackletMatching()
+{
+  const auto& param = NA6PRecoParam::Instance();
+  const auto& layout = NA6PLayoutParam::Instance();
+  auto& clusters = getClusters();
+
+  std::vector<NA6PTrack> trks = mMSTracker->getTracks();
```
shahor02: A few more comments: here the full copy of tracks is created, while there is no real need for that: better to change the NA6PTrackerCA getter to  `std::vector<NA6PTrack>& getTracks()` and do here 
```
const auto& trks = mMSTracker->getTracks();
```
The follow-up comment is below.

rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -149,7 +163,170 @@ void NA6PMuonSpecReconstruction::closeTracksOutput()
...
+  for (size_t i = 0; i < clusters.size(); ++i) {
+    int originalID = clusters[i].getClusterIndex();
+    if (originalID >= 0 && (size_t)originalID < clusters.size())
+      clusterLookup[originalID] = static_cast<int>(i);
+  }
+
+  // matching loop
+  std::vector<MatchCandidate> candidates;
+  for (int jT = 0; jT < nTrks; jT++) {
+    NA6PTrack tr = trks[jT];
```
shahor02: Here, another full copy of the MS track is created, but it is enough to take `const auto& tr = trks[jT];`
The the outer param of this track is propagated for each tracklet through a lot of material in small steps. Even if we assume that eventually the tracklets of the same layer may start at different Zs, it is better to propagate once a copy to some Z in the vicinity of the 1st MID plane, then to propagate its copy to every tracklet for much shorter distance. 
Besides that, whenever the cov. matrix is not used, it is much faster to propagate only NA6PTrackPar part only.

rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -149,7 +163,170 @@ void NA6PMuonSpecReconstruction::closeTracksOutput()
...
+      float xyzClu2[3] = {clu2.getX(), clu2.getY(), clu2.getZ()};
+
+      if (xyzClu1[2] > xyzClu2[2]) {
+        float tmp[3] = {xyzClu1[0], xyzClu1[1], xyzClu1[2]};
+        for (int j = 0; j < 3; ++j) {
+          xyzClu1[j] = xyzClu2[j];
+          xyzClu2[j] = tmp[j];
+        }
+      }
+      float zToProp = xyzClu1[2];
```
shahor02: Checking the findTracklets method, I don't see how z1 > z2 is possible, I think this check can be omitted.

rec/src/NA6PMuonSpecReconstruction.cxx
fprino: Aren't the next access to the vector faster in the tracker if the vector itself is sorted and we access contiguous elements in memory? 
For instance, we use:
    auto layerBegin = cluArr.begin() + firstIndex[iLayer + 1];
    auto layerEnd = cluArr.begin() + lastIndex[iLayer + 1];
and 
      auto lower = std::partition_point(layerBegin, layerEnd,
 
Sorting the indices, would have a a consequence that in the next steps we access scattered locations in memory. Could this slow down the tracking?

For this reason, the solution of storing the original index seemed to me the easiest and cheapest.

What about adding these protections:
https://github.com/shahor02/NA6PRoot/compare/main...fprino:NA6PRoot:protectCluIndex?expand=1
The idea behind them is:
1) initialize mClusterIndex to -1, which can later be used to tag cases in which the cluster index is not set
2) call setClusterIndex also inside hitsToRecPoints (and digitsToRecPoints) so that the index is assigned also to the clusters stored in the "owned" mCluster array.
It seems to me that this could address the fragility spotted by codex

shahor02: In this particular case, the std::partition_point will profit from the contiguous clusters sorted in an optimal way, but only provided they fit to processor cache (which likely will not happen once we consider real N clusters). 
But even in this case, if you want to profit from the locality, then it is better to work with smaller objects than the clusters, exactly to fit them to the cache. E.g. sort not the vector of clusters, but `std::vector<std::pair<tan2, clIndex>>`, with the tan2 being the value used for the partition_point lambda.

In all other cases we are anyway jumping from cluster of one layer to that of another.

rec/src/NA6PMuonSpecReconstruction.cxx
```diff
@@ -149,7 +163,170 @@ void NA6PMuonSpecReconstruction::closeTracksOutput()
...
+    refitInw.setOuterParam(outTr);
+    refitInw.resetClusters();
+    fitter->addClustersToTrack(refitInw);
+    if (param.vtDoConstrainedTrack) {
+      fitter->constrainTrackToVertex(refitInw, *mPrimaryVertex);
+    } else {
+      refitInw.getVertexConstrainedParam().invalidate();
+    }
+    refitInw.setStatusMS(NA6PTrack::kMSMatchedToMIDRefitted);
+    mTracks.push_back(refitInw);
```
shahor02: Considering the comments above, I've tested this version (for lines 185-330):

```
  const auto& trks = mMSTracker->getTracks();

  auto saveMSOnly = [&trks, this](int j, int flag) {
    this->mTracks.push_back(trks[j]);
    this->mTracks.back().setStatusMS(flag);
  };

  int nTrks = trks.size();
  int firstMID = layout.nVerTelPlanes + layout.nMSPlanes - 2;
  std::vector<std::pair<NA6PMuonSpecCluster, NA6PMuonSpecCluster>> trkltsMID = mMSTracker->findTracklets(firstMID, firstMID + 1, clusters, mPrimaryVertex);
  int nTrklets = trkltsMID.size();
  NA6PFastTrackFitter* fitter = mMSTracker->getTrackFitter();

  // create lookup table of cluster indices
  std::vector<int> clusterLookup(clusters.size(), -1);
  for (size_t i = 0; i < clusters.size(); ++i) {
    int originalID = clusters[i].getClusterIndex();
    if (originalID >= 0 && (size_t)originalID < clusters.size())
      clusterLookup[originalID] = static_cast<int>(i);
  }

  // matching loop
  std::vector<MatchCandidate> candidates;
  std::vector<NA6PTrackParCov> outTrProp; // to reuse propagated outer tracks
  if (nTrklets) {
    for (int jT = 0; jT < nTrks; jT++) {
      const auto& tr = trks[jT];
      auto& trOutCopy = outTrProp.emplace_back(tr.getOuterParam());
      if (!Propagator::Instance()->propagateToZ(trOutCopy, param.msZForMSMIDmatch, fitter->getPropOpt())) {
        continue;
      }
      for (int jS = 0; jS < nTrklets; jS++) {
	const auto& tracklet = trkltsMID[jS];
	NA6PLine trkLine(tracklet.first.getXYZ(), tracklet.second.getXYZ());
	NA6PTrackPar trOutCopyT{trOutCopy};
	if (!Propagator::Instance()->propagateToZ(trOutCopyT, trkLine.mOriginPoint[2], fitter->getPropOpt())) {
	  continue;
	}
	auto trackDirCos = trOutCopyT.getDirCos();
	float distXY = std::hypot(trkLine.mOriginPoint[0]-trOutCopyT.getX(),  trkLine.mOriginPoint[1]-trOutCopyT.getY());
	float costh = NA6PLine::getScalar(trkLine.mCosinesDirector, trackDirCos);
	if (distXY < param.msMaxDistTrackMSTrackletMID && costh > param.msMinCosThetaTrackMSTrackletMID) {
	  costh = std::clamp(costh, -1.f, 1.f);
	  float score = distXY / 2. + (1 - costh) / 1.e-4;
	  candidates.push_back({jT, jS, score});
	}
      }
    }
  }
  std::sort(candidates.begin(), candidates.end(), [](auto& a, auto& b) { return a.score < b.score; });
  std::vector<bool> trackUsed(nTrks, false);
  std::vector<bool> segmentUsed(nTrklets, false);
  std::vector<int> trackMatch(nTrks, -1);

  for (auto& c : candidates) {
    if (trackUsed[c.track])
      continue;
    if (segmentUsed[c.segment])
      continue;
    trackMatch[c.track] = c.segment;
    trackUsed[c.track] = true;
    segmentUsed[c.segment] = true;
  }

  for (int jT = 0; jT < nTrks; jT++) {
    int jS = trackMatch[jT];
    if (jS < 0) { // track not matched to MID: save with specific status flag
      saveMSOnly(jT, NA6PTrack::kMSNotMatchedToMID);
      continue;
    }
    auto& trOut = outTrProp[jT]; // propagated tracks are guaranteed to exist if there was at least 1 match (or even tracklet)
    const auto& tracklet = trkltsMID[jS];
    fitter->cleanupAndStartFit();
    fitter->addCluster(tracklet.first);
    fitter->addCluster(tracklet.second);    
    if (fitter->fitSeedOutward(trOut, false)<0) { // track not refitted: save with specific status flag. 
      saveMSOnly(jT, NA6PTrack::kMSMatchedToMIDnotRefitted);
      continue;
    } // RS: do we want to check the chi2 of this refit with MID clusters?
    
    // add MS clusters for refit inward
    const auto& trMS = trks[jT];
    for (int jl = 0; jl < 4; ++jl) {
      // RS: avoid this complications by using directly orig cluster indices in the tracklets      
      int originalID = trMS.getClusterIndex(jl + param.vtNLayers);
      if (originalID >= 0 && (size_t)originalID < clusterLookup.size()) {
        int jNewPos = clusterLookup[originalID];
        if (jNewPos < 0) {
          LOGP(error, "Cluster originalID={} not found for track {} layer {}", originalID, jT, jl);
          continue;
        }
        const auto& cl = clusters[jNewPos];
        fitter->addCluster(cl);
      }
    }
    auto trInw = trOut;
    float chi2Refit = fitter->fitSeedInward(trInw, true);
    if (chi2Refit < 0.f) { // track not refitted: save with specific status flag
      saveMSOnly(jT, NA6PTrack::kMSMatchedToMIDnotRefitted);
      continue;
    }
    saveMSOnly(jT, NA6PTrack::kMSMatchedToMIDRefitted);
    auto& trMSFin = mTracks.back();
    trMSFin.setChi2(chi2Refit);
    trMSFin.setOuterParam(trOut);
    ((NA6PTrackParCov&)trMSFin) = trInw;
    trMSFin.resetClusters();
    fitter->addClustersToTrack(trMSFin);
    if (param.vtDoConstrainedTrack) {
      fitter->constrainTrackToVertex(trMSFin, *mPrimaryVertex);
    } else {
      trMSFin.getVertexConstrainedParam().invalidate();
    }
  }
```

Note that I had to fix a bug in the NA6PLine

rec/src/NA6PMuonSpecReconstruction.cxx
shahor02: Thinking more about this: actually, I am not sure that in general case the sorting in eta per layer and using partition_point to find needed range is optimal. Eventually we will need to consider the same set of clusters with different vertex constraints, sorting for each of them will be too consuming. Also, using eta for the matches for the clusters e.g. in the vicinity of X,Y, we  are probing all combinations of `|X|,|Y|`. Would not it be easier to sort once e.g. in Y, then, for every vertex constraint, search (e.g. with the partition_point) for the matching Y range. 
In this way we will be working in the native cluster coordinates (at most shifted by vertex position) instead of some transform function, like eta or its aliases. The same is true for discarding X-regions outside of the search regions.

fprino:
Hi Ruben,
thanks for the suggestion and for sharing your improved code.
It is actually great to use methods from NA6PTrack and NA6PLine to compute the track-tracklet distance and pointing.
I adopted essentially all your modifications in the commit that I just pushed.
I tested it on my 5000 J/psi events: compared to the previous version, the fraction of MS tracks that are not matched did not change (as expected). The fraction of tracks that failed in the refit when adding the MID points increased by a tiny bit, not sure why. I will do some further checks later.

shahor02:
Thanks @fprino 
Just the last change: to avoid the intermediate copy of the tracks from the TrackerCA, which are then anyway replicated to MID matched tracks, I've added a direct refernce gettter of mFinalTracks. I've opened a PR https://github.com/fprino/NA6PRoot/pull/4 to you branch.

fprino:
I merged your PR, thanks!

PR #126, opened 2026-07-15
Fix uncertainties in secondary vertex fitting

galocco:
This PR: 
- Removes the dummy uncertainty on z and uses the track slopes to compute it
- Fix the second-order track derivatives
- Uses the full covariance matrix in the vertex fit
- Implements createParentTrackParCov
- Add option to use all 3 components of the B field (important for long-lived particles)
- Add pull, strange particles, and fix the mass computation in the test macro

For the reconstruction of the decay vertices of long-lived particles, vtPropagateTracksToPV should be set to false (the default is true).

### DIFF
--- a/rec/include/NA6PDCAFitterN.h
+++ b/rec/include/NA6PDCAFitterN.h
@@ -1,22 +1,30 @@
 #ifndef _NA6P_DCA_FITTERN_H
 #define _NA6P_DCA_FITTERN_H
 
+#include <algorithm>
+#include <cmath>
+
 #include <Math/SVector.h>
 #include <Math/SMatrix.h>
 #include <fairlogger/Logger.h>
 #include "NA6PTrack.h"
 #include "NA6PHelixHelper.h"
+#include "Propagator.h"
 
 using CircleXZ = NA6PHelixHelper::CircleXZ;
 
 struct TrackCovI {
-  float sxx, syy, sxy, szz;
+  // Independent elements of the symmetric 3D information matrix
+  // H^T Cxy^{-1} H. A track constrains X and Y at a given Z through
+  // H = {{1, 0, -dx/dz}, {0, 1, -dy/dz}}.
+  float sxx, syy, sxy, sxz, syz, szz;
 
-  TrackCovI(const NA6PTrack& trc, float zerrFactor = 1.f) { set(trc, zerrFactor); }
+  TrackCovI(const NA6PTrack& trc) { set(trc); }
   TrackCovI() = default;
-  bool set(const NA6PTrack& trc, float zerrFactor = 1.f)
+  bool set(const NA6PTrack& trc)
   {
-    float cxx = trc.getSigmaX2(), cyy = trc.getSigmaY2(), cxy = trc.getSigmaYX(), czz = cyy * zerrFactor;
+    // Invert the 2D covariance of the measured track position (X,Y).
+    float cxx = trc.getSigmaX2(), cyy = trc.getSigmaY2(), cxy = trc.getSigmaYX();
     float detXY = cxx * cyy - cxy * cxy;
     bool res = true;
     if (detXY <= 0.) {
@@ -28,7 +36,11 @@ struct TrackCovI {
     sxx = cyy * detXYI;
     syy = cxx * detXYI;
     sxy = -cxy * detXYI;
-    szz = 1. / czz;
+    const float sx = trc.getSlopeX();
+    const float sy = trc.getSlopeY();
+    sxz = -(sxx * sx + sxy * sy);
+    syz = -(sxy * sx + syy * sy);
+    szz = sx * sx * sxx + 2.f * sx * sy * sxy + sy * sy * syy;
     return res;
   }
 };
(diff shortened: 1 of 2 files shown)


galocco:
@shahor02 @fprino I was wondering if it would be better to remove vtPropagateTracksToPV from the configurables.

rec/include/NA6PDCAFitterN.h
```diff
@@ -696,7 +753,15 @@ void NA6PDCAFitterN<N, Args...>::calcTrackDerivatives()
...
-    mTrDer[mCurHyp][i].set(mCandTr[mCurHyp][i], mBy);
+    const auto& trc = mCandTr[mCurHyp][i];
+    float by = mBy;
+    if (mUseFullFieldPropagation) {
+      const auto* propagator = Propagator::Instance();
+      if (propagator->hasMagFieldSet()) {
+        by = propagator->getBy(trc.getXYZ());
+      }
+    }
+    mTrDer[mCurHyp][i].set(trc, by);
```
shahor02: Not sure it is worth doing this in the loop: this function is called for the tracks already propagated to the PCA seed, so the trc.getXYZ should be almost the same for all tracks. It should be enough to query the field from the 1st track before the loop. Also, it does not worth to check if the field is set in the frequently called methods: it must be set in advance, if it is not, then let the job crash and we fix it. I would do something like:
```
float by = mUseFullFieldPropagation ? Propagator::Instance()->getBy(mCandTr[mCurHyp][0].getXYZ()) : mBy;
for (int i = N; i--;) {
  mTrDer[mCurHyp][i].set(trc, by);
}
```

rec/include/NA6PDCAFitterN.h
```diff
@@ -968,17 +1038,26 @@ NA6PTrack NA6PDCAFitterN<N, Args...>::createParentTrackParCov(int cand) const
...
-      covV[MomInd[i]] += covT[MomInd[i]];
-    }
+    const double pxz = trc.getPxz();
+    const double tx = trc.getTx();
+    const double q2pxz = trc.getQ2Pxz();
+    const double cosPsi = trc.getCosPsi();
+    const double dPxdQ = -pvecT[0] / q2pxz;
+    const double dPydQ = -pvecT[1] / q2pxz;
+    const double dPzdTx = -pxz * tx / cosPsi;
+    const double dPzdQ = -pvecT[2] / q2pxz;
```
shahor02: These 3 divisions would be optimized out by the compiler to 1 div. and 3 multiplications only with -Ofast compilation option: https://godbolt.org/z/n5xhqs5bP
Better to not rely on it and do directly
```
const double pxzI = 1./pxz;
...
const double dPxdQ = -pvecT[0] * pxzI;
const double dPydQ = -pvecT[1] * pxzI;
const double dPzdQ = -pvecT[2] * pxzI;
```

rec/include/NA6PDCAFitterN.h
```diff
@@ -991,8 +1070,45 @@ NA6PTrack NA6PDCAFitterN<N, Args...>::createParentTrackParCov(int cand) const
   covV[3] = covVtxV(2, 0);
   covV[4] = covVtxV(2, 1);
   covV[5] = covVtxV(2, 2);
-  NA6PTrack tr; // TODO: update adding covariance matrix
+  const double px = pvecV[0], py = pvecV[1], pz = pvecV[2];
```
shahor02, line 1066: I did not follow the derivations below, if you have a codex or chatGPT page for it, could you share it please? It might be worth considering it also for the O2.

shahor02 (commented):
Thanks @galocco, see a few comments below.
For the vtTracks propagation to DCA: I agree that it is better to keep the track at its innermost update, like in O2. Moreover, the final PV is not known during the VT tracking

rec/include/NA6PDCAFitterN.h
galocco: With Codex, I generated the documentation containing the derivation of the parent-track covariances: [[NA6PDCAFitterN-parent-covariance.pdf](https://github.com/user-attachments/files/30084218/NA6PDCAFitterN-parent-covariance.pdf)](https://github.com/user-attachments/files/30084218/NA6PDCAFitterN-parent-covariance.pdf). The calculation seems correct.

I checked the parent-track pulls for (D^+ \rightarrow K\pi\pi) and (\Lambda^0 \rightarrow p\pi), using `mUseAbsDCA=false`, and they look good:

### D+ -> Kpipi

Dplus-TrackPulls

### Lambda0 -> ppi

Lambda0-ParentTrackPulls

rec/include/NA6PDCAFitterN.h
galocco: You're right. Sampling the magnetic field for each track has no effect. I've implemented your suggestion.

rec/include/NA6PDCAFitterN.h
galocco: Interesting, I didn’t know that. From now on, I’ll pay more attention to this kind of optimization. I’ve implemented your suggestion.

rec/include/NA6PDCAFitterN.h
shahor02: Thanks a lot! I'll propagate them to O2.

shahor02 (approved):

